PR #158, opened 2026-04-09
feat: anonymous usage telemetry

abacigil:
## Summary

- Add privacy-respecting, opt-out anonymous telemetry to understand how DecisionBox is used
- Enabled by default, disabled with `TELEMETRY_ENABLED=false` or `DO_NOT_TRACK=1`
- Events batched hourly (configurable via `TELEMETRY_FLUSH_INTERVAL`) and sent to a Cloudflare Worker + D1
- Public API key (`X-API-Key` header) filters non-DecisionBox traffic, per-install rate limiting (100 events/hr)

### What's collected (all anonymous)
- Install ID (random UUID), version, OS/arch, deployment method
- Provider types (warehouse, LLM), domain
- Bucketed counts and durations (never exact values)
- Error classes (never error messages)

### What's never collected
- PII, credentials, query content, table names, insight content

### Implementation
- `libs/go-common/telemetry/` — client library with batching, 16 unit tests
- API: `server_started`, `server_stopped`, `project_created` events
- Agent: `discovery_completed`, `discovery_failed` events with bucketed metrics
- Cloudflare Worker: [decisionbox-telemetry-worker](https://github.com/decisionbox-io/decisionbox-telemetry-worker) (separate repo, deployed to `telemetry.decisionbox.io`)
- `TELEMETRY.md` — full documentation of every event and field
- `docs/reference/telemetry.md` — docs site reference page
- `setup.sh` — telemetry opt-out prompt during setup wizard
- Updated: CHANGELOG, README, docker-compose.yml, configuration.md

### RFC Discussion
- #157

## Test plan

- [x] `make build` passes
- [x] `make test-go` passes (16 new telemetry tests)
- [x] `make lint-go` passes (0 issues)
- [x] Telemetry disabled: no events sent, no network requests (`TELEMETRY_ENABLED=false`)
- [x] Telemetry disabled via `DO_NOT_TRACK=1`
- [x] Events arrive at Cloudflare D1 (verified end-to-end with custom domain)
- [x] Requests without `X-API-Key` are rejected (403)
- [x] Rate limiting works (429 after 100 events/hr per install_id)
- [ ] Verify graceful shutdown flushes events (`kill -SIGTERM <pid>`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Anonymous usage telemetry** -- Collects anonymous, privacy-respecting usage metrics (version, OS, provider types, event counts). Enabled by default, disable with `TELEMETRY_ENABLED=false` or `DO_NOT_TRACK=1`. No PII, no query content, no credentials. See [TELEMETRY.md](TELEMETRY.md) for full details.
+
 ### Changed
 
 - **Domain packs are now dynamic** -- stored in MongoDB instead of compiled Go code. Create, edit, import, and export domain packs from the dashboard without code changes.
--- a/README.md
+++ b/README.md
@@ -190,8 +190,13 @@ Key environment variables:
 | `SECRET_PROVIDER` | `mongodb` | Secret storage: `mongodb`, `gcp`, `aws` |
 | `RUNNER_MODE` | `subprocess` | Agent runner: `subprocess`, `kubernetes` |
 | `LLM_TIMEOUT` | `300s` | Timeout per LLM API call |
+| `TELEMETRY_ENABLED` | `true` | Anonymous usage telemetry ([details](TELEMETRY.md)) |
 
 Full reference: [Configuration](https://decisionbox.io/docs/reference/configuration).
+
+## Telemetry
+
+DecisionBox collects anonymous usage telemetry to help improve the product. No PII, query content, or credentials are ever collected. Disable with `TELEMETRY_ENABLED=false` or `DO_NOT_TRACK=1`. See [TELEMETRY.md](TELEMETRY.md) for full details.
 
 ## Documentation
 
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -162,7 +162,8 @@ The telemetry implementation is fully open source:
 - **API integration**: [`services/api/apiserver/apiserver.go`](services/api/apiserver/apiserver.go) -- init, server events
 - **Agent integration**: [`services/agent/agentserver/agentserver.go`](services/agent/agentserver/agentserver.go) -- discovery events
 
-Events are batched in memory and sent every hour (or on shutdown).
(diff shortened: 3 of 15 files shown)


osalkk (changes requested):
Good feature — privacy approach, documentation, and test coverage are solid. One blocking issue before merge:

## Blocking: Shutdown race condition

`shutdown()` in `telemetry.go` calls `flush()` which spawns `go c.send(batch)`, then returns immediately. The final batch's HTTP request runs in a detached goroutine that may be orphaned when the process exits. This is the scenario in your unchecked test plan item ("Verify graceful shutdown flushes events").

**Fix**: Add a `sync.WaitGroup` to track in-flight `send()` goroutines, and call `wg.Wait()` in `shutdown()` before closing the done channel.

## Non-blocking suggestions

1. **`log.Printf` in `install_id.go:34`** — codebase requires structured logging (`apilog`/`applog`), not stdlib `log`. Since this is go-common, consider silently dropping the error or accepting a logger interface.
2. **`TELEMETRY_FLUSH_INTERVAL` undocumented** — the env var is used and tested but not in `configuration.md` or `TELEMETRY.md`.
3. **Version var duplicated** in both `agentserver.go` and `apiserver.go` — should be shared or set via `-ldflags`.
4. **`deploymentMethod()` docker-compose detection** — checking `MONGODB_URI == "mongodb://mongodb:27017"` is fragile. Consider `/.dockerenv` file existence.
5. **Test `time.Sleep` for sync** — could be flaky under CI load. Channel-based signaling would be more robust.

abacigil:
All feedback addressed in 36f42ec:

**Blocking:**
- **Shutdown race fixed** — Added `sync.WaitGroup` to track in-flight `send()` goroutines. `shutdown()` now calls `wg.Wait()` before closing the done channel, ensuring the final batch is fully sent before process exit.

**Non-blocking:**
1. **stdlib `log` removed** — All `log.Printf` calls replaced with silent drops (both `telemetry.go` and `install_id.go`). go-common no longer imports `log`.
2. **`TELEMETRY_FLUSH_INTERVAL` documented** — Added to `configuration.md` and `TELEMETRY.md`.
3. **Version deduplicated** — Moved to `libs/go-common/version/version.go`, imported by both API and agent. Supports `-ldflags` override at build time.
4. **Docker detection fixed** — Uses `/.dockerenv` file existence instead of fragile `MONGODB_URI` string comparison.
5. **Flush interval default changed** — From `1h` to `5m`.

osalkk (approved):
All feedback addressed cleanly in 36f42ec:

- **Shutdown race**: Fixed with `sync.WaitGroup` — `wg.Wait()` in `shutdown()` ensures in-flight sends complete ✅
- **stdlib log removed**: Errors silently dropped, no more `log.Printf` in go-common ✅
- **`TELEMETRY_FLUSH_INTERVAL` documented**: Added to configuration.md and TELEMETRY.md ✅
- **Version deduplicated**: New `libs/go-common/version/` shared package, both services import it ✅
- **Docker detection fixed**: Uses `/.dockerenv` stat check ✅

CI green. LGTM.

PR #161, opened 2026-04-10
chore(terraform/aws): remove orphaned ip_allowlist security group

abacigil:
## Summary

setup.sh switched to \`alb.ingress.kubernetes.io/inbound-cidrs\` for AWS IP allowlisting in #155. With that annotation, the AWS Load Balancer Controller stays in its default "I manage backend SG rules" mode and the standalone \`aws_security_group.ip_allowlist\` resource is no longer attached to anything — it's dead code created on every \`terraform apply\` and never used.

This PR removes it.

## What it removes

- \`aws_security_group.ip_allowlist\` (and its 3 supporting \`aws_security_group_rule\` resources for HTTP/HTTPS ingress + egress)
- \`ip_allowlist_security_group_id\` output (both the module-level output in \`modules/decisionbox/outputs.tf\` and the proxied output in \`prod/outputs.tf\`)
- Stale comment block in \`vpc.tf\` that referenced the old \`alb.ingress.kubernetes.io/security-groups\` attachment pattern (replaced with a short note pointing at the inbound-cidrs approach)
- Outdated description for \`allowed_ip_ranges\` in \`docs/deployment/terraform-aws.md\` (which still said "Creates a security group for ALB attachment")

## What it does NOT change

- \`var.allowed_ip_ranges\` is still the variable users set. setup.sh still consumes it and passes it to the helm chart as \`inbound-cidrs\`.
- The GCP \`google_compute_security_policy.ip_allowlist\` (Cloud Armor) is unrelated and stays.

## Why this is safe

The resource was created on \`terraform apply\` whenever \`allowed_ip_ranges\` was non-empty, but nothing in the helm chart, the module, or setup.sh has referenced its ID since #155 landed. Any deployment that already migrated to the inbound-cidrs flow will see the SG removed on the next \`terraform apply\` with no functional impact.

For deployments that are still on the OLD security-groups annotation flow: they need to migrate the helm release to inbound-cidrs FIRST (remove the \`security-groups\` annotation, add \`inbound-cidrs\`), then \`terraform apply\` to clean up the SG. Otherwise the apply will fail trying to delete a SG that's still attached to an in-use ALB.

## Validation

- \`terraform validate\` clean on \`terraform/aws/modules/decisionbox\`
- \`terraform validate\` clean on \`terraform/aws/prod\`
- \`grep -rn 'ip_allowlist' terraform/aws docs/\` shows zero residual references in the AWS tree (all remaining hits are GCP Cloud Armor)

## Test plan

- [ ] CI green
- [ ] On a fresh AWS install (\`./setup.sh --cloud aws\` with \`allowed_ip_ranges\` set), verify the ALB ingress is reachable from the allowed CIDR and rejected from outside, and that no orphaned SG exists in the account
- [ ] On the existing internal cluster, the helm release is migrated to inbound-cidrs first, then \`terraform apply\` cleanly removes the unused SG

### DIFF
--- a/docs/deployment/terraform-aws.md
+++ b/docs/deployment/terraform-aws.md
@@ -226,7 +226,7 @@ Set `region` and `cluster_name` to match your environment.
 | `secret_namespace` | string | `decisionbox` | Secret name prefix for IAM scoping |
 | `enable_bedrock_iam` | bool | `false` | Grant Bedrock InvokeModel access (Agent) |
 | `enable_redshift_iam` | bool | `false` | Grant Redshift Data API read access (Agent) |
-| `allowed_ip_ranges` | list(string) | `[]` | CIDR blocks allowed for HTTP/HTTPS. Empty = unrestricted. Creates a security group for ALB attachment. |
+| `allowed_ip_ranges` | list(string) | `[]` | CIDR blocks allowed for HTTP/HTTPS. Empty = unrestricted. Applied via the `alb.ingress.kubernetes.io/inbound-cidrs` annotation. |
 
 ### Tags
 
--- a/terraform/aws/modules/decisionbox/outputs.tf
+++ b/terraform/aws/modules/decisionbox/outputs.tf
@@ -59,8 +59,3 @@ output "redshift_iam_enabled" {
   description = "Whether Redshift IAM was enabled"
   value       = var.enable_redshift_iam
 }
-
-output "ip_allowlist_security_group_id" {
-  description = "Security group ID for IP allowlisting (empty if no IP restriction). Attach to ALBs via alb.ingress.kubernetes.io/security-groups annotation."
-  value       = length(var.allowed_ip_ranges) > 0 && var.create_vpc ? aws_security_group.ip_allowlist[0].id : ""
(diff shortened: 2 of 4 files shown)


osalkk (approved):
Verified against the live AWS cluster:

- `setup.sh` on main only sets `inbound-cidrs`, no `security-groups` annotation
- Helm chart templates and stored release values have no `security-groups` reference
- The Terraform SG resource is orphaned — safe to remove

Code changes are clean: removes the dead SG + 3 rules, cleans up outputs, updates docs. LGTM.

**Note for apply**: the live ingress still has a leftover `security-groups: sg-07400b511b7bf9b0d` annotation from a manual kubectl debug session. Remove it before running `terraform apply`, otherwise AWS will refuse to delete the SG while it's attached to the ALB ENIs:

```bash
kubectl annotate ingress ingress-decisionbox-dashboard -n decisionbox \
  alb.ingress.kubernetes.io/security-groups- \
  alb.ingress.kubernetes.io/manage-backend-security-group-rules-
```

Wait for the ALB controller to detach the SG (~30s), then apply.

PR #168, opened 2026-04-16
fix(api): propagate QDRANT_URL and QDRANT_API_KEY to agent Jobs

abacigil:
## Summary

- Agent discovery Jobs were missing `QDRANT_URL` and `QDRANT_API_KEY` env vars. The agent's `initQdrant()` returned nil, the orchestrator's `vectorStore` was nil, and `phase_embed_index` was **silently skipped** on every run — leaving insights saved to MongoDB but never indexed in Qdrant.
- This broke the **Ask** feature and project-scoped **semantic search** for every project running under Kubernetes mode. Ask's RAG retrieval returned zero matches and fell through to the "No relevant insights found" fallback message.
- The API pod itself has `QDRANT_URL` set via the Kubernetes service — the value just wasn't being forwarded when spawning agent Jobs in `services/api/internal/runner/kubernetes.go`.

## Fix

Two lines added to the env-var propagation list in `buildJob`:

```go
{"QDRANT_URL", "QDRANT_URL"},
{"QDRANT_API_KEY", "QDRANT_API_KEY"},
```

Uses the same conditional-propagation pattern as the existing `SECRET_*` vars: only set on the Job if set on the API process. Unset → absent from Job spec (not empty string, which would look "configured" but broken).

## Tests

Added two unit tests in `runner_test.go` using the existing fake K8s client setup:

- `TestKubernetesRunner_Run_PropagatesQdrantEnv` — sets both env vars, runs a Job, asserts both land in the container spec with correct values.
- `TestKubernetesRunner_Run_OmitsQdrantEnvWhenUnset` — asserts they do NOT appear in the Job env when unset on the API process.

Both pass locally. Full `./internal/runner/...` suite green. `golangci-lint` clean.

## Impact

- Any existing discoveries on this or similar clusters already have un-embedded insights. Operators can run `decisionbox-api backfill-embeddings --project-id <id>` in the API pod to catch up. After this PR is deployed, new runs self-index as they complete.
- No config changes needed on the API side — `QDRANT_URL` is already set in the API pod env.
- Helm chart values (`values.yaml`) already document `QDRANT_URL` on the API deployment; no chart changes required.

## Test plan

- [x] `make test-go` (runner package) — passes
- [x] `make lint-go` (runner package) — 0 issues
- [x] Build clean on services/api
- [ ] Verify in internal-prod cluster after merge: next discovery run should see `QDRANT_URL` in its pod env (`kubectl describe pod discovery-*`), and insights should appear in Qdrant without manual backfill

### DIFF
--- a/services/api/internal/runner/kubernetes.go
+++ b/services/api/internal/runner/kubernetes.go
@@ -72,6 +72,8 @@ func (r *KubernetesRunner) buildJob(spec jobSpec) *batchv1.Job {
 		{"SECRET_NAMESPACE", "SECRET_NAMESPACE"},
 		{"SECRET_ENCRYPTION_KEY", "SECRET_ENCRYPTION_KEY"},
 		{"SECRET_GCP_PROJECT_ID", "SECRET_GCP_PROJECT_ID"},
+		{"QDRANT_URL", "QDRANT_URL"},
+		{"QDRANT_API_KEY", "QDRANT_API_KEY"},
 	} {
 		if v := getEnv(kv.envKey, ""); v != "" {
 			envVars = append(envVars, corev1.EnvVar{Name: kv.key, Value: v})
--- a/services/api/internal/runner/runner_test.go
+++ b/services/api/internal/runner/runner_test.go
@@ -284,6 +284,67 @@ func TestKubernetesRunner_Cancel_NotFound(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


osalkk (approved):
Clean, minimal fix — two lines in the right place, follows the existing conditional-propagation pattern exactly. Tests cover both the presence and absence cases. PR description is thorough with clear root cause analysis and operator backfill instructions.

Ship it.

PR #174, opened 2026-04-22
fix(api): ship decisionbox-agent binary inside API image

abacigil:
## Summary

- The API's default `RUNNER_MODE=subprocess` spawns the agent via `exec.Command(\"decisionbox-agent\", ...)`, but `services/api/Dockerfile` only shipped `decisionbox-api`. Following the Docker Compose Quick Start and starting a discovery failed with `exec: \"decisionbox-agent\": executable file not found in $PATH`.
- The API Dockerfile now also builds and installs `decisionbox-agent` into `/usr/local/bin`, so `docker compose up -d` works end-to-end out of the box.
- Kubernetes deployments are unaffected — they use `RUNNER_MODE=kubernetes` and run the agent as a Job from its own image. The standalone `services/agent/Dockerfile` is unchanged.

## Test plan

- [x] \`docker build -f services/api/Dockerfile .\` succeeds
- [x] \`which decisionbox-api\` and \`which decisionbox-agent\` both resolve inside the image
- [x] \`decisionbox-agent --help\` runs and shows the flags \`SubprocessRunner\` passes (\`--project-id\`, \`--run-id\`, \`--areas\`, \`--max-steps\`)
- [ ] Full Quick Start: \`docker compose up -d\` → create project → run discovery → completes without the PATH error


erenbaydemir (approved):
lgtm

PR #194, opened 2026-04-30
fix(dashboard): raise API proxy timeout to 20 minutes for long-running endpoints

abacigil:
## Summary

Next.js middleware uses Node's global \`fetch\` (undici), whose default \`headersTimeout\` and \`bodyTimeout\` are both **5 minutes**. That cap kills long synchronous endpoints — most visibly pack-generation synth on Opus, which routinely runs 5–8 minutes against a 270k-char prompt. The Go API has no 5-min timeout of its own; the dashboard proxy in front of it does.

\`middleware.ts\` now imports \`undici.fetch\` with a shared \`Agent\` whose timeouts are raised to **20 minutes**, and pins itself to the Node.js runtime so the dispatcher option is honoured. Browser-side fetch has no comparable timeout, so raising the proxy is sufficient end-to-end.

Long-term, pack-gen should move to an async-job model (return 202 + run_id, poll project state) and eliminate the synchronous HTTP path entirely. This fix unblocks the synchronous path until that lands.

## Why the Go API isn't the fix point

Trace: \`browser → dashboard middleware proxy → Go API → Bedrock\`. The dashboard's \`await fetch(targetUrl, ...)\` is itself the bottleneck. When undici's 5-min default fires, it tears down the upstream socket; Go sees the client disconnect and cancels \`r.Context()\`, which propagates to the AWS SDK's \`InvokeModel\` mid-flight. That's the \"context canceled, StatusCode: 0, RequestID: empty\" pattern in the field report — duration always exactly \`5m0.69xs\`. The Go API's \`WriteTimeout=30s\` doesn't cancel the handler context; only the dashboard proxy gates request lifetime end-to-end.

## Test plan

New jest suite \`src/__tests__/proxyTimeout.test.ts\` boots a slow stub HTTP server and asserts:

1. **Short-Agent rejects with \`UND_ERR_HEADERS_TIMEOUT\`** — proves the dispatcher's \`headersTimeout\` option is the actual gate, not just a knob that does nothing.
2. **20-minute Agent (the value middleware ships) waits for a 3-second response** — proves the long timeout takes effect.
3. **Timeout constant stays ≥ 15 minutes** — guard rail; an accidental edit dropping back to undici's 5-minute default fails this.

Verified locally:
- [x] \`npm run test\` — 165 dashboard tests pass (162 prior + 3 new)
- [x] \`npm run lint\` — clean
- [x] \`npm run build\` — green
- [x] Manual run of the same logic against a slow server: 6/6 contract assertions pass

## Notes for review

- \`undici\` added as an explicit dashboard dependency (Node 22 doesn't expose it as a builtin yet — that's Node 23+).
- \`jest.setup.ts\` now guards on \`window\` / \`Element\` so the new \`@jest-environment node\` test can run without breaking the existing jsdom suite.
- Next.js 16 deprecation warning suggests renaming \`middleware.ts\` to \`proxy.ts\`; deferred to a separate PR (different concern).


abacigil:
@copilot review the latest commit (9e7d27e) — addresses all 5 findings:

1. **Body cast** (`middleware.ts:60`): replaced `as unknown as undefined` with a typed cast to `UndiciBodyInit | undefined` and `request.body ?? undefined` so the body type is honest. Runtime behaviour was already correct (cast doesn't affect runtime), but the type annotation no longer lies.
2. **Request headers** (`middleware.ts:58`): pass the `Headers` instance directly via `as unknown as UndiciHeadersInit` — multi-value headers (Cookie repeats) preserved.
3. **Response headers** (`middleware.ts:66`): construct `new Headers(response.headers as unknown as HeadersInit)` directly — repeated `Set-Cookie` values preserved.
4. **Test cleanup** (`proxyTimeout.test.ts:37`): `SlowServer.close` now returns a Promise that resolves after the http.Server's close callback fires, awaited in every teardown path.
5. **Engines pin** (`package.json`): added `engines.node: ">=20.18.1"` so older Node 20 patches fail install loudly.

Plus a new regression test that asserts a POST body round-trips through undici fetch + the long-running dispatcher (4th case, total 166 tests pass).

PR #184, opened 2026-04-26
feat(agent): on-demand schema retrieval to fix 1M-token context exhaustion

abacigil:
## Summary

- The exploration agent used to dump a Level-1 schema block (column lists + 3 sample rows for the top-K retriever matches) into the system prompt of every step. On wide warehouses (~2K tables) this combined with the per-step SQL result accumulation to hit the Bedrock 1M-token limit at ~step 98 (`prompt is too long: 1002763 tokens > 1000000 maximum`), killing healthy discovery runs.
- This PR replaces the upfront L1 dump with two on-demand actions the LLM can call when it actually needs detail: `lookup_schema` (up to 10 fully-qualified refs per call, served from the in-memory schemas cache; no warehouse traffic) and `search_tables` (semantic Qdrant query when the catalog hint isn't enough). Per-run budgets: 30 lookups, 30 searches; topK clamped to 30.
- `{{SCHEMA_INFO}}` is now the single canonical schema variable and renders the Level-0 catalog only. The `{{SCHEMA_CATALOG}}` and `{{SCHEMA_RETRIEVED}}` placeholders introduced earlier in this release cycle are gone. Telemetry: per-action counters `schema_lookup_calls` and `schema_search_calls` replace the old `schema_inspect_table_calls`.

## What's in the change

**Production code**
- `SchemaProvider` interface with `Lookup` / `Search` and run-level constants
- `CacheSchemaProvider` (in-memory schemas + Qdrant retriever) with qualified / bare-name / case-insensitive ref resolution and ambiguity detection
- Action parser + budget enforcement in `ExplorationEngine`, with per-call dedup against `fetchedTables` to prevent budget waste on repeats
- Per-action telemetry counters; new `lookup_schema` / `search_tables` exploration step types
- Dead `inspect_table.go` removed (was never wired into production)
- All four domain-pack exploration prompts (gaming, social, ecommerce, system-test) rewritten with the new action contract
- Token budget math in `estimate.go` updated for the 70/20/10 query/lookup/search mix

**Tests**
- `exploration_actions_test.go`: parseAction shapes (incl. legacy), normaliseRefs, formatters, executeLookupSchema and executeSearchTables across success / dedup / partial dedup / per-call cap / budget exhausted / provider error / no provider / topK clamp / topK default, engine wiring defaults, end-to-end scripted run (lookup → search → query → done)
[...]

PR #195, opened 2026-04-30
fix(agent): cap schema-cache sample rows to fit Mongo 16MB BSON limit

abacigil:
## Summary

\`SchemaCacheRepository.Save\` was failing silently on wide-row warehouse tables (VARCHAR(MAX) / XML / JSON / wide nested-struct columns) because a single 5-row sample doc exceeded Mongo's hard 16 MB BSON limit. The indexer logged the failure as a warning and continued, leaving \`schema_index_status: \"ready\"\` while \`project_schema_cache\` was empty for the project — every subsequent discovery run then errored with \`schema cache lookup failed: schema cache is empty for this project\` and \`/reindex\` did nothing because the same overflow re-occurred on every retry.

Save now caps cached sample rows at 3 per table and truncates each string / \`[]byte\` value at 256 chars before insertion. The truncation marker \`…(truncated, original N chars)\` keeps downstream consumers honest — blurb generation and the on-demand schema lookup see the value is incomplete so the truncated form isn't matched against as the real value.

## Field repro

Oxxo POC, project \`69f34eddc7e6887b16b4ccce\` (MSSQL DW, dbo schema):

\`\`\`
WARN  schema_indexer: schema-cache save failed; next run will rediscover
err=\"schema cache save: an inserted document is too large\"
\`\`\`

Combined with \`schema_index_status: \"ready\"\` and \`db.project_schema_cache.countDocuments({project_id: ...}) === 0\` — exactly what triggered this PR.

## Why the indexer doesn't fail loudly

This is intentional — see the comment block in \`resolveSchemas\`:

> The cache is strictly best-effort: Find errors degrade to a fresh discovery (logged + continue) and Save errors don't fail the run (the next run just rediscovers).

The bug was that the next run would always hit the same overflow, so the cache stayed empty forever and discovery had no fallback to live schema discovery (it's strict-cached now since #184). This PR makes the Save itself reliable so the soft-failure doesn't matter.

## Why caps over schema redesign

Three caps over redesigning the cache shape because:

- **Row cap (3)** — blurb generation only needs a few examples to capture value shape. The cache_schema_provider already caps further to 3 by default before serving to the LLM, so storing 5+ in cache wastes bytes nobody reads.
- **Value cap (256 chars)** — long enough to convey shape (URL prefix, format, key fragment, JSON head); short enough that even a 200-column row marshals to <100 KB.
- **Truncation marker** — matches the format the LLM already understands as \"abridged\" so it doesn't fabricate matches against the truncated tail.

## Test plan

- [x] 7 new unit tests in \`schema_cache_repo_test.go\` cover: row-count cap, per-value truncation marker, byte-slice handling, non-string pass-through, input non-mutation, and an adversarial 1000-column × 6-row × oversized-value case asserting the capped output stays well under 16 MB.
- [x] \`make test-go\` (full Go suite) — green.
- [x] \`make lint-go\` — 0 issues.
- [ ] After merge: rebuild enterprise images, pull on POC, re-run reindex, confirm \`db.project_schema_cache.countDocuments({project_id: \"69f34eddc7e6887b16b4ccce\"})\` is non-zero, then re-run discovery.


abacigil:
@copilot review the latest commit (c9bd898) — addresses all 4 findings:

1. **Multi-byte rune split** (`schema_cache_repo.go`): `capSampleString` now counts runes via `utf8.RuneCountInString` and slices on a rune boundary by walking with `range`. New regression test `TestCapSampleString_NeverSplitsMultibyteRune` asserts the truncated output is valid UTF-8 on a 257-rune Turkish "ı" string.

2. **Non-UTF-8 `[]byte`** (`schema_cache_repo.go`): `capSampleValue` now uses `utf8.Valid([]byte)` to route — UTF-8 bytes go through the text path, non-UTF-8 (VARBINARY / BLOB) goes through new `capSampleBinary` which base64-encodes the prefix. New marker `…(truncated binary, original N bytes, base64)` distinguishes binary from text. Regression tests cover both the under-limit base64 round-trip and the over-limit truncated case.

3. **Test column key collision** (`schema_cache_repo_test.go`): keys now include `strconv.Itoa(i)` so the adversarial case has 1000 distinct columns. Test also asserts the column count after capping matches the input.

4. **CHANGELOG marker mismatch** (`CHANGELOG.md`): documented both markers (text/rune and binary/base64) with the rationale.

13 unit tests total, all pass. `make test-go` + `make lint-go` clean.

PR #196, opened 2026-05-01
fix(agent): ground insight verifier on cited source-step SQL (Layer 1)

abacigil:
## Summary

Layer 1 of the verification-grounding fix designed in [`plans/PLAN-INSIGHT-VERIFICATION-GROUNDING.md`](https://github.com/decisionbox-io/decisionbox-plans/blob/main/PLAN-INSIGHT-VERIFICATION-GROUNDING.md). The insight verifier now renders the SQL of every step the analyst LLM cited as `source_steps` into its generation prompt as priority-1, column-grounded evidence — ranked above the catalog, with a strict instruction to prefer column names that already executed successfully against this warehouse rather than invent new ones.

### Why

Customer report 2026-04-30 against an MSSQL Netsis-style warehouse: 9 of 10 insights ended with `validation.status = "error"` and `Invalid column name 'TARIiH' / 'STHAR_SUBE' / 'SUBEKODU' / …`. The verifier received only a table-level catalog (no column names) and hallucinated columns when it had no other context. The exploration phase, which already had a `lookup_schema` tool loop, executed cleanly against the same warehouse — only verification produced the "Invalid column name" errors.

This PR closes that gap for the common case (single-table insights where `source_steps` covers the columns the verifier needs to count). Layers 2 (per-call `FixOpts` for the self-healing SQL fixer), 3 (verifier-side `lookup_schema` loop for cross-table cases), and 4 (same grounding for `user_count_validator`) follow in subsequent PRs.

### What changed

- **New package** `services/agent/internal/validation/render` — `RenderVerificationContext` shared between this PR's verifier prompt and the upcoming PR-2 fixer prompt. `DefaultBudgetChars = 32_000`, oldest-step-first drop policy under budget pressure (per plan §4.1 v2 update — narrowest, most recent step has the highest signal).
- **`InsightValidator` now requires `SetExplorationLog`** — calling `ValidateInsights` without it panics. No backward compat: this is a wiring bug, not a runtime condition (plan §1.1).
- **Orchestrator wires** `explorationResult.Steps` into the validator after the exploration phase completes and before the analysis loop runs.
- **Prompt update** — source-queries section + grounding rule rendered above the catalog; the catalog stays as supplementary context for tables source steps didn't touch.
- **Docs lockstep** — `docs/concepts/discovery-lifecycle.md` Phase 6 rewrite explaining source-step grounding; `docs/architecture/agent-on-demand-schema.md` extended with a Verification-phase column-grounding section noting Layer 1 is shipped, Layers 2/3 are planned.
- **CHANGELOG** entry under `[Unreleased]`.

### Test plan

- [x] `go test ./...` passes across the agent module (validation, render, discovery, ai, etc.)
- [x] `golangci-lint run ./...` clean across services/agent + libs + every provider module
- [x] New unit tests in `internal/validation/render/`:
    - nil/empty inputs return empty
    - cited step not in log → silently skipped
    - non-`query_data` steps (`lookup_schema`, `search_tables`) → skipped
    - empty / whitespace-only `Query` → skipped
    - duplicate citations de-duped
    - citation order preserved
    - multi-line SQL fenced cleanly, trailing newlines normalized
    - error step renders error annotation, not row count
    - missing `QueryPurpose` falls back to `(no purpose recorded)`
    - budget cap drops oldest step number first (regardless of citation order)
    - zero budget → `DefaultBudgetChars`
    - rule instruction wording regression guard
- [x] New unit tests in `internal/validation/insight_validator_layer1_test.go`:
    - `ValidateInsights` panics if `SetExplorationLog` was never called
    - `SetExplorationLog(nil)` panics
    - Source SQL of a cited step (`STHAR_TARIH`) appears in the verification prompt
    - Empty `SourceSteps` ⇒ section header omitted (legacy behavior preserved for now; PR-3 replaces with `lookup_schema` fallback)
    - Stray citation tolerated
    - `lookup_schema` step skipped from rendering even when cited
    - Budget pressure: newest step survives
    - Rule instruction lands AFTER the rendered evidence
- [ ] Real-LLM + real-warehouse integration suite + soak test (per plan §5.4 / §5.4.1) — deferred to a follow-up that lands with PR-3, when the cross-table cases are real and the integration surface is complete. PR-1 alone fixes the customer's reported single-table failure mode.
- [ ] Customer remediation against `192.1.0.157` — deferred until PR-1+PR-2 are merged and `:latest` images are published (plan §7.2).

### Self-review

Walked the diff after writing tests:
- Removed the legacy `ONLY use column names that exist in the table schemas above` clause when introducing the new rule, then realized that regressed the no-source-steps fallback path. Restored a catalog-aware rule in the `CRITICAL TABLE NAME RULES` block so both paths stay grounded. Diff caught by re-reading `generateVerificationQuery`.
- Confirmed the validator's `Validation` block is the only field written — `insight.AffectedCount` is left intact. Documented in the lifecycle doc so future readers don't assume a quietly destructive change.


abacigil:
Addressed all four Copilot comments in 33c99d3:

- **render.go package comment** — removed the customer IP. Kept the generic Netsis-style description + Turkish column example for the WHY breadcrumb.
- **`trimToBudget` O(n²) → O(n log n)** — precompute per-step rendered size once (single `writeStep` per step into a scratch builder), keep a running `total` and decrement it on each drop. The sort dominates now; no more whole-section re-rendering inside the drop loop.
- **`bySize` → `byStepAsc`** — the sort is by `Step` number (age), so the new name reflects the drop-policy intent.
- **`SetExplorationLog` doc comment** — reworded to match the implementation: the validator retains the slice and does not copy elements / backing array, so callers must not mutate it afterwards.

Tests + lint clean locally; all required CI checks were green on `0d00674` and the fixes are doc/refactor-only — no behavioral change.

PR #197, opened 2026-05-01
feat(agent): per-call FixOpts in SQL fixer with verification context (Layer 2)

abacigil:
## Summary

Layer 2 of the verification-grounding fix designed in [`plans/PLAN-INSIGHT-VERIFICATION-GROUNDING.md`](https://github.com/decisionbox-io/decisionbox-platform/blob/main/plans/PLAN-INSIGHT-VERIFICATION-GROUNDING.md). When the verifier's SQL fails with `Invalid column name`, the self-healing fixer now sees the same source-step grounding evidence the verification prompt was built on — so on retry it can substitute the real column name rather than re-emit the original hallucination. Builds on PR #196 (Layer 1).

## What changed

**API (no backward compat per plan §1.1):**
- `queryexec.SQLFixer.FixSQL` now takes a `queryexec.FixOpts` argument carrying per-call grounding context.
- `queryexec.QueryExecutor.ExecuteWithFixOpts(ctx, query, purpose, opts)` is the new entry point. `Execute(ctx, query, purpose)` is a thin shim that forwards `FixOpts{}` — exploration callers stay zero-opts; only the validator's `executorAdapter` calls `ExecuteWithFixOpts` with non-empty opts.
- `validation.SelfHealingExecutor` adds the FixOpts argument; the validator computes `RenderVerificationContext(insight.SourceSteps)` per insight and forwards it.

**Prompt template:**
- New `applySection` helper in `services/agent/internal/ai/sql_fixer.go` implements mustache-style `{{#NAME}}…{{/NAME}}` conditional sections. When the var is empty (after trimming) the entire block — markers + inner content — is removed; when populated, the markers are stripped and the inner content is kept.
- All six warehouse `prompts/sql_fix.md` (BigQuery, Redshift, Snowflake, Postgres, Databricks, MSSQL) gain a `{{#VERIFICATION_CONTEXT}}…{{/VERIFICATION_CONTEXT}}` block with shared phrasing of the column-grounding rule. Each warehouse can specialize the wording without forking the whole prompt.

## Test plan

- [x] `go test ./...` clean across services/agent and all six warehouse provider modules.
- [x] `golangci-lint run ./...` clean across services/agent + 6 warehouse provider modules.
- [x] `applySection` table tests cover: kept/removed/whitespace-only/multiple-occurrences/no-markers/unterminated-marker/different-named-section.
- [x] `TestSQLFixer_FixSQL_VerificationContextSection{Kept,StrippedWhenEmpty}` assert the per-call FixOpts both populates the conditional block and triggers section removal.
- [x] `validation.captureExecutor` asserts the validator forwards the rendered VerificationContext into FixOpts when SourceSteps is non-empty (`STHAR_TARIH` shows up via the `RenderVerificationContext` output) and forwards an empty VerificationContext when SourceSteps is empty.
- [x] Each warehouse `provider_test.go` asserts `{{#VERIFICATION_CONTEXT}}`, `{{VERIFICATION_CONTEXT}}`, `{{/VERIFICATION_CONTEXT}}` are present in the embedded `sql_fix.md` so a future template edit can't silently strip the layer's evidence.
- [x] `query_executor_test.go` uses a local `mockFixer` (replaces the testutil-level mock) to break the `queryexec_test → testutil → queryexec` import cycle that emerges when `FixOpts` is referenced from testutil.
- [ ] Real-LLM + real-warehouse integration / soak suite — bundled with PR-3 (Layer 3) when the cross-table cases are real.

## Self-review

- Walked the diff after writing tests. Confirmed:
  - The `Execute` shim still receives the same security-filter check inside `ExecuteWithFixOpts` (forwarding only the new opts; no logic moved out).
  - `validation.SelfHealingExecutor` change ripples to the orchestrator's `executorAdapter` (now uses `ExecuteWithFixOpts`) and the `TestExecutorAdapter*` tests; the latter were updated to pass `queryexec.FixOpts{}`.
  - `MockSQLFixer` removed from `testutil/mocks.go` because keeping it there would force `testutil → queryexec`, creating the queryexec-test cycle. The local replacement in `query_executor_test.go` is the only consumer.
  - All warehouse template inserts are identical phrasing — easy to spot if a future prompt edit drifts. The provider_test.go assertions catch silent removal.


abacigil:
Addressed all four Copilot comments in e6d671c:

- **insight_validator_layer1_test.go** — `TestInsightValidator_ForwardsEmptyFixOpts_WhenInsightHasNoSourceSteps` now asserts `exec.calls == 1` so the empty-VerificationContext check is meaningful. Without that, an early validator exit could pass the test silently.
- **sql_fixer_test.go** — `applySection` table-test failure message now includes the section name argument so failures show all three inputs.
- **query_executor_test.go** — two new focused tests:
  - `TestExecuteWithFixOpts_ForwardsOptsToFixerOnRetry` asserts the validator's rendered `VerificationContext` reaches the fixer on retry.
  - `TestExecute_ShimAlwaysPassesEmptyFixOpts` pre-populates the mock fixer with a sentinel `LastOpts` to confirm the legacy `Execute` shim cannot leak per-call grounding state to the explore path.
- **snowflake provider_test.go** — required-placeholder list now includes `{{FILTER}}` and `{{CONVERSATION_HISTORY}}` (already present in the postgres/redshift/databricks/mssql provider tests).

Tests + lint clean locally; doc/test-only refactor — no behavioral change.

PR #198, opened 2026-05-01
feat(agent): verifier lookup_schema tool loop (Layer 3)

abacigil:
## Summary

Layer 3 of the verification-grounding fix designed in [`plans/PLAN-INSIGHT-VERIFICATION-GROUNDING.md`](https://github.com/decisionbox-io/decisionbox-platform/blob/main/plans/PLAN-INSIGHT-VERIFICATION-GROUNDING.md). The insight verifier is no longer a single-shot LLM call. With a `SchemaProvider` wired, it runs up to `MaxLookupsPerVerification = 6` `lookup_schema` rounds per insight before emitting the final SELECT COUNT — letting the model fetch column detail for tables that `source_steps` did not cover. Builds on PR #196 (Layer 1) and PR #197 (Layer 2).

### Why

Layer 1 covered single-table verifications where `source_steps` already touches the columns the verifier needs to count. For cross-table insights (a JOIN across two tables, only one of which appeared in source steps) Layer 1 alone left the model guessing. Layer 3 closes that gap by giving the verifier its own `lookup_schema` access — same provider the explorer already uses, same cache, no live warehouse traffic.

### What changed

**Shared action parser** (plan §6.6 — drift prevention):
- `(e *ExplorationEngine).extractJSON` lifted to a free function (the body never used receiver state).
- `ai.ParseAction(response, allowed []string)` is the new package-level entry point. `nil` allowed → any action (the explorer's full set). The verifier passes `["lookup_schema", "query_data"]` so the model can't "complete" mid-verify and explorer-only future actions don't leak in.
- `(e *ExplorationEngine).parseAction` stays as a thin wrapper passing `nil` — existing exploration tests untouched.

**Verifier loop** (`services/agent/internal/validation/insight_validator.go`):
- `runVerificationLoop` drives the LLM through up to `MaxLookupsPerVerification` rounds. After each `lookup_schema` action the SchemaProvider's tables are merged into a deduplicated accumulator and rendered into the next round's prompt + the final `FixOpts`.
- When the budget is exhausted before a query, **one forced final round** (NOT counted against the budget) asks for raw SQL. If that still doesn't yield a query, the validator returns `validation.status="error"` with `reasoning="exhausted lookup budget"` rather than reusing an arbitrary fallback.
- Without a `SchemaProvider`, the verifier falls through to `generateSingleShotQuery` (Layer 1 + 2 only).

**Render extension** (`services/agent/internal/validation/render/render.go`):
- New `RenderVerificationContextWithLookups` extends `RenderVerificationContext` with on-demand lookup detail. The lookup section follows the source-queries block — Layer 1 evidence is more current than the cache (per plan §6.10).

**Orchestrator wiring**:
- After building the shared `CacheSchemaProvider` for the explorer, the orchestrator forwards the same instance to the validator via `SetSchemaProvider`.

### Test plan

- [x] `go test ./...` clean across services/agent.
- [x] `golangci-lint run ./...` clean.
- [x] `internal/validation/insight_validator_layer3_test.go` covers (per plan §5.3):
    - Lookup served in one round (verify `FixOpts.VerificationContext` contains the looked-up table + columns).
    - Budget exhausted → forced final round emits SQL (verify count of provider lookups, ≤ MaxLookupsPerVerification).
    - Forced round still no query → status="error" with reasoning="exhausted lookup budget".
    - NotFound surfaced cleanly (model self-corrects on next round).
    - No SchemaProvider → single-shot fallback (assert exactly 1 LLM call, no tool loop).
    - Lookup results land in FixOpts.VerificationContext alongside source-step rendering.
    - Allow-list rejects `complete`; accepts `lookup_schema`/`query_data`.
    - Combined source + lookup evidence in FixOpts when both are present.
- [ ] Real-LLM integration / soak harness — deferred per the plan's collapse condition (§7.1). All three layers now ship; an integration suite with the customer's column-shape variants is the natural next PR.

### Self-review

Walked the diff after writing tests. Confirmed:
- The free-function `extractJSON` body matches the previous receiver method exactly — only the receiver was dropped.
- `parseAction` (the method wrapper) still calls `ParseAction(response, nil)`, so the explorer's behaviour is unchanged. Two existing exploration tests that constructed an empty `*ExplorationEngine` only to call `extractJSON` were updated to call the free function directly.
- `validateSingleInsight`'s `vr.Reasoning` field now interpolates the underlying error so the "exhausted lookup budget" telemetry signal is preserved through to the validation result (was previously a hardcoded "Could not generate" message that swallowed the cause).
- `mergeLookups` deduplicates by qualified table name — a model that re-asks for a table it already has won't get duplicate entries in the rendered evidence block.
- `bareSQLFallback` accepts a raw SELECT response inside the loop too — matches the LLM-behavior reality where the model sometimes ignores the JSON envelope rule and just emits the query.


abacigil:
Addressed all three Copilot comments in 8e4c5cf:

- **`ParseAction` doc comment** — corrected the verifier allow-list to `{"lookup_schema", "query_data"}` (the rest of the codebase, tests, and PR description already reflected the smaller set).
- **`LookupResult.NotFound` / `Truncated` now fed back into the prompt** — the loop accumulates both across rounds and `buildVerificationPrompt` renders a "Lookup Notices" section so the model can self-correct (retry a misspelled `dataset.table`, stop re-requesting a known-missing ref, follow up on a truncated batch). New `appendUniqueStrings` helper dedupes the NotFound accumulator across rounds. Single-shot callers pass `(nil, false)` so no notices section appears for them.
- **Test comment fix** — `TestInsightValidator_LookupSchemaBudgetExceeded` previously implied request-level dedup at the SchemaProvider; rewritten to clarify that `mergeLookups` dedupes only the rendered accumulator and the upstream `Lookup` is called every round.

New regression test: `TestInsightValidator_NotFoundFedBackToModel` walks the second-round prompt and asserts both the not-found ref and the "Lookup Notices" header land in the prompt, so a regression that drops the surfacing path is caught.

Tests + lint clean locally.

PR #199, opened 2026-05-01
feat(agent): user-count probe carries grounded FixOpts on retry

abacigil:
## Summary

The `UserCountValidator`'s hardcoded `COUNT(DISTINCT user_id)` probes used to fail silently on warehouses where the user-id column isn't named `user_id` — every insight's `affected_count` claim went through unverified. The probe now routes through the self-healing executor with `FixOpts` carrying the source-step grounding evidence, so the SQL fixer can substitute the real column name (e.g. `KULLANICI_ID`, `customer_id`, `USER_ID_HASH`) on retry.

Builds on PR #196, PR #197, PR #198.

### Why

After the verifier-side grounding work, the insight verifier finally produces grounded SELECT COUNT queries. But the `UserCountValidator`'s separate user-count probe (used as a sanity-check denominator for `affected_count`) was still issuing the same hardcoded `COUNT(DISTINCT user_id) FROM sessions` against the raw warehouse client. On warehouses with non-English user-id columns this probe returned `Invalid column 'user_id'`, the validator fell to "could not determine total users", and the orchestrator skipped its sanity check entirely. This PR closes that gap.

### What changed

**API:**
- `UserCountValidatorOptions` gains an optional `Executor` field (the same `SelfHealingExecutor` the `InsightValidator` uses).
- New `SetExplorationLog(log)` and `SetExecutor(exec)` methods follow the orchestrator's construct-then-set pattern (the validator is built in `New()` before `RunDiscovery`'s executor exists).

**Probe path:**
- `runProbeQuery` dispatches each probe through `executor.Execute` when one is wired, falling back to `v.warehouse.Query` otherwise.
- Each executor call carries `FixOpts.VerificationContext` rendered from the **union of all `query_data` step IDs** in the exploration log (run-wide evidence, since the probe is single-shot per run rather than per-insight).
- New helpers: `extractTotalUsersFromRows` centralises the count extraction; `collectQueryStepIDs` filters non-`query_data` steps out of the source-step ID list.

**Orchestrator wiring:**
- After `executor` is built in `RunDiscovery`, the orchestrator calls `userCountValidator.SetExecutor` with the same `executorAdapter` the `InsightValidator` uses.
- The existing `SetExplorationLog` block now wires both validators with the same `explorationResult.Steps` slice.

### Test plan

- [x] `go test ./...` clean across services/agent.
- [x] `golangci-lint run ./...` clean.
- [x] `user_count_validator_layer4_test.go` covers:
    - `SetExplorationLog(nil)` panics (mirrors the `InsightValidator` contract).
    - Probe routes through executor when wired.
    - `FixOpts.VerificationContext` carries source-step columns + step numbers (Turkish-column shape: `KULLANICI_ID`, `STHAR_TARIH`).
    - Falls through to direct `warehouse.Query` when no executor (legacy path preserved).
    - Empty exploration log → empty `VerificationContext` (executor still receives a valid FixOpts).
    - `collectQueryStepIDs` filters lookup_schema/search_tables steps.
    - `extractTotalUsersFromRows` handles missing field, non-numeric values, all integer types.
    - `SetExecutor` replaces a previously-wired executor (orchestrator construct-then-set wiring).
- [x] Existing `user_count_validator_test.go` untouched — the no-executor fallback path matches legacy behaviour, all existing tests pass without modification.

### Self-review

- The `Executor` is optional on the options struct; existing tests + the orchestrator's `New()` construction (where the executor doesn't yet exist) both stay valid by passing `nil`. The setter pattern matches the `InsightValidator`.
- `collectQueryStepIDs` returns an empty slice (not nil) for an empty log — `RenderVerificationContext` is happy with either, but explicit empty slice keeps the behaviour deterministic.
- `runProbeQuery`'s fallback to direct warehouse access matches the prior behaviour exactly (same `extractTotalUsersFromRows`, same null-safety).
- The orchestrator's wiring change is paired: `SetExecutor` is called early (before exploration), `SetExplorationLog` is called late (after exploration). Either alone is a runtime no-op in terms of verification — both are needed to enable the self-healing grounded probe.


abacigil:
Addressed all five Copilot comments in ca32a7e:

- **`SetExplorationLog` docstring** — the "MUST be called" framing was wrong; `GetTotalUsers` works fine without it (empty VerificationContext is just... empty). Softened to "optional; empty log produces no grounding evidence and the SQL fixer falls back to its standard prompt".
- **`collectQueryStepIDs` whitespace handling** — switched from `s.Query == ""` to `strings.TrimSpace(s.Query) == ""` for consistency with `render.isExecutableQueryStep`.
- **Lazy `fixOpts` rendering** — `RenderVerificationContext` is now only called when `v.executor != nil`. The no-executor path skips the work entirely.
- **Case-insensitive `total_users` extraction** — Snowflake (and other warehouses that fold unquoted aliases) returns `TOTAL_USERS`. `extractTotalUsersFromRows` now does `strings.EqualFold` against the column key, so the extractor works portably across BigQuery / Snowflake / Postgres without forcing dialect-specific aliases.
- **Cache invalidation on setter swap** — both `SetExecutor` and `SetExplorationLog` now clear `totalUsersCached` so a swap mid-run actually takes effect. Without this, the orchestrator's late-binding of the executor in `RunDiscovery` would silently keep returning whatever was computed via the prior path.

Five new regression tests cover each fix. Tests + lint clean locally.

PR #200, opened 2026-05-01
feat(agent,api,ui): split discovery logs into per-step collections

abacigil:
## Summary

A 97-step exploration on a wide warehouse blew past Mongo's 16MB-per-document limit on the first save attempt:

```
Failed to save discovery result: an inserted document is too large
```

The `discoveries` collection embedded every LLM dialog as an array on the parent doc (`exploration_log`, `analysis_log`, `validation_log`, `recommendation_log`). The `discovery_runs` collection had the same unbounded-growth problem in its embedded `steps` array, which `StatusReporter` `\$push`'d row-by-row during the run.

This PR splits each log type into its own Mongo collection — one row per step / area / validation / RunStep, keyed by the parent discovery's `_id` (or the `run_id` for live status). The dashboard rehydrates them through paginated endpoints rather than dragging the parent doc.

## What changed

**New collections:**
- `discovery_exploration_steps` (one per `ExplorationStep`)
- `discovery_analysis_steps` (one per `AnalysisStep`)
- `discovery_validation_results` (one per `ValidationResult`)
- `discovery_recommendation_log` (one per discovery, unique on `discovery_id`)
- `discovery_run_steps` (one per live `RunStep`)

**Writers (agent):**
- `DiscoveryLogRepository.Save{ExplorationSteps,AnalysisSteps,ValidationResults,RecommendationLog}` persists each log type after the parent doc save. Partial failures log but do not roll back the discovery.
- `RunStepRepository.AddStep` replaces the `\$push` on the run document. One `InsertOne` per step.
- `StatusReporter` takes the new repo + `projectID`; `RunRepository.AddStep` and the `Steps []RunStep` slice initialiser are gone.

**Readers (api):**
- `DiscoveryLogRepository` (read-only) backs five new endpoints:
  - `GET /api/v1/discoveries/{id}/exploration-steps[?limit=N]`
  - `GET /api/v1/discoveries/{id}/analysis-steps`
  - `GET /api/v1/discoveries/{id}/validation-results`
  - `GET /api/v1/discoveries/{id}/recommendation-log`
  - `GET /api/v1/runs/{runId}/steps[?since=<RFC3339>&limit=N]`
- The `DiscoveryRepo.List` projection that filtered out the now-removed embedded fields is gone.

**Models:**
- `DiscoveryResult` drops `ExplorationLog` / `AnalysisLog` / `ValidationLog` / `RecommendationLog` (both agent and api copies).
- `DiscoveryRun` drops `Steps` (both copies).
- JSON/BSON round-trip tests assert the legacy fields are absent from the wire format (regression guard).

**Dashboard:**
- `DiscoveryResult` interface drops the embedded log fields.
- New `api.listExplorationSteps` / `listAnalysisSteps` / `listValidationResults` / `getRecommendationLog` / `listRunSteps` fetchers.
- Discovery-detail and insight-detail pages fetch the three log types via the new endpoints in parallel with the parent doc; each is independently optional (a failed split-log fetch doesn't blank the page).

**No backward compatibility:** DecisionBox is pre-1.0. Runs persisted before this change keep their embedded fields as orphan data the new endpoints don't surface; new runs use the split collections from the moment the upgraded image rolls out.

## Test plan

- [x] `go test ./...` clean across `services/agent` and `services/api` (with the existing JSON round-trip tests rewritten to assert the legacy log fields are NOT in the wire format).
- [x] `golangci-lint run ./...` clean across both modules.
- [x] `services/agent/internal/database/discovery_log_repo_integration_test.go` (build tag `integration`) round-trips each log type against a Mongo testcontainer:
  - Round-trip with multiple steps / areas / validations / one recommendation log.
  - Empty-input no-op contract on every Save method.
  - Per-discovery isolation: rows for one discovery never leak into reads for another.
  - Limit honoured.
- [x] `services/agent/internal/database/run_step_repo_integration_test.go` covers `since`-cursor streaming, run-isolation, timestamp auto-fill.
- [x] `services/api/internal/handler/discoveries_split_logs_test.go` unit-tests each new endpoint with mock repos: happy path, missing path-value, nil repo, invalid `since` cursor, limit clamping at 5000, repo error → 500.
- [x] Dashboard `tsc --noEmit` clean (the unrelated `undici` errors on `proxyTimeout.test.ts` and `middleware.ts` predate this PR).

## Self-review

- Walked the diff after writing tests. Confirmed:
  - The agent's `DiscoveryRepository.Save` no longer touches log fields (they're gone from the struct), so the 16MB scenario is structurally impossible — the parent doc is bounded by `insights` + `recommendations` + `summary` + `schemas`, all of which are structured outputs.
  - The status-reporter's per-step `\$push` is replaced with `InsertOne`. The legacy `RunRepository.AddStep` is removed entirely (no dead code shim).
  - `agentserver.go` wires both new repos and runs `EnsureIndexes` on each at startup. `apiserver/server.go` wires the read-side repos and registers the five new routes (gated `viewer` like the rest of the discovery endpoints).
  - The split-collection rows carry `project_id` + `discovery_id` (or `run_id`) + `created_at` so cross-collection queries work without a join. Recommendation-log index is unique on `discovery_id`.
  - The dashboard's two pages that rendered embedded logs (`page.tsx` and `insights/[insightId]/page.tsx`) are updated to fetch from the new endpoints. Each split-log call is wrapped in `.catch(() => {})` so a single failed fetch doesn't blank the whole page.
  - Pre-existing tests with hardcoded `NewDiscoveriesHandler(...)` / `NewStatusReporter(...)` call shapes were updated to the new arg counts. No tests skipped.
- Architecture doc at `docs/architecture/discovery-log-collections.md` describes the write/read paths, indexes, and migration stance.

## Follow-up (next PR)

`UserCountValidator.GetTotalUsers` still issues three hardcoded `COUNT(DISTINCT user_id) FROM <dataset>.{sessions,events,app_users}` probes. On warehouses without those exact tables / column conventions, the SQL fixer adapts them on retry but at high token cost. Next PR replaces the hardcoded probes with a `lookup_schema`-driven LLM tool loop (same shape as `InsightValidator.runVerificationLoop` from #198), grounding the probe on source-step evidence + on-demand schema lookups.


abacigil:
Addressed all three Copilot comments + a duplication concern in 41db3ff:

- **`RecommendationLogEntry.RunAt`** — was `string`; agent persists `run_at` as BSON datetime. Switched to `time.Time` (JSON still emits RFC3339, so the dashboard wire shape is unchanged). Without this every GET on the recommendation-log endpoint would have 500'd.
- **`OrchestratorOptions.DiscoveryLogRepo` doc comment** — claimed Required-with-panic but the persistence call sites guard with `if != nil`. Doc updated to match the actual optional behaviour. Production builds always wire it; the nil branch is for unit tests that don't bring up MongoDB.
- **`ListExplorationSteps` limit** — now clamped to `maxExplorationStepsPerRequest = 1000`, mirroring the bound on `GetDebugLogs`. Exploration step rows are large (full LLM dialog + query results); leaving the limit unbounded defeats the point of the split.
- **Duplication audit** — collection name strings were repeated between `services/agent/internal/database/mongodb.go` (exported constants) and the api side's local `collection*` constants. Hoisted them into a new `libs/go-common/mongodb/collections.go`; both services import from there. The agent re-exports them through its `database` package so existing `database.Collection*` call sites stay compilable.

All three modules (`libs/go-common`, `services/agent`, `services/api`) build + test + lint clean locally.

abacigil:
Addressed all three Copilot round-2 findings in 6bc5581 + audited each as a bug pattern across the branch:

1. **`ListRunSteps` unbounded `limit`** — `limit <= 0 || limit > max` is the right shape, mirroring `ListExplorationSteps`. Defined `maxRunStepsPerRequest = 5000`, added regression tests for missing-limit and negative-limit defaults. Audit confirmed no other endpoint in this PR has the same bug.

2. **Hardcoded collection name in `services/api/database/run_step_repo.go`** — switched to `gomongo.CollectionDiscoveryRunSteps`. Audit confirmed every other writer/reader on both sides goes through `libs/go-common/mongodb` constants now.

3. **`discovery_recommendation_log` cardinality comments** — reconciled the package header and the `RecommendationLogDoc` struct comment with what the index + writer actually do (exactly one doc per discovery, unique on `discovery_id`, `InsertOne` once per discovery save).

**Plus a fourth issue Copilot did not flag, found while auditing for the same pattern:** the orchestrator stored `discoveryLogRepo` as an interface but assigned it directly from the concrete `*DiscoveryLogRepository` option. A typed-nil pointer through that path would produce a non-nil interface with a nil concrete pointer — the `o.discoveryLogRepo == nil` guard in `persistSplitLogs` would then return false and dereference the nil. Same Go gotcha I had to fix on `runStepWriter` last commit. Mirrored the normalization in `NewOrchestrator` and added a regression test that constructs with a typed-nil option and asserts the field is normalized.

All three modules build + test + lint clean locally.

abacigil:
Round-3 Copilot fixes in d8e632a, plus a fourth bug the audit caught:

1. **`$gt` ms-precision timestamp cursor (both repos)** — real correctness bug. Two AddStep calls inside the same BSON millisecond would produce same-`timestamp` rows; the dashboard's last-rendered timestamp would then `$gt`-skip any later row that fell in that ms, silently dropping step rows from the live panel on long runs. Switched the cursor to an opaque ObjectID hex (`?since=<hex>` from a prior response). ObjectIDs increment per writer process via their internal counter, so collisions are impossible. Added an integration test that drives three inserts at the *same* explicit ms timestamp and pages across the boundary — pre-fix that test would have lost a row. Index switched from `(run_id, timestamp)` to `(run_id, _id)`. Repo returns `database.ErrInvalidCursor` on a malformed hex; handler maps that to 400 (caller input), not 500.

2. **`discovery_log_repo` header comment lied about indexes** — header said "indexed by (project_id, discovery_id, ...)" but EnsureIndexes actually builds `(discovery_id, step/run_at/validated_at)` for the list views and a separate `(project_id, created_at desc)` only on the exploration collection. Replaced the comment with the real per-collection breakdown.

**Plus a fourth issue the audit caught that Copilot missed:** `LiveRunPanel` in `ui/dashboard/src/app/projects/[id]/page.tsx` was reading `run.steps`, but this PR removed `steps[]` from `DiscoveryRun`. The live-run panel was effectively rendering an empty list. Wired it to `api.listRunSteps(run.id, lastID)` with the new opaque cursor — polling at 1.5s while the run is live, one final fetch on terminal status. Updated the TS `RunStep` interface to add `id` and updated the fetcher to take the opaque cursor instead of an RFC3339 timestamp.

All three modules build + test + lint clean locally. Dashboard `tsc --noEmit` is clean for the changed files (the only remaining errors are pre-existing `undici` stubs unrelated to this PR).

abacigil:
Round-4 fixes in 12c5139 — pure doc drift from the cursor-shape switch. Updated `docs/architecture/discovery-log-collections.md` (collections table + read-path section) and `CHANGELOG.md` to describe the opaque ObjectID cursor and the rationale (ms-collision risk). Audited the rest of the tree; only `debug-logs` docs reference RFC3339 `since`, which is correct — different endpoint, not affected by this fix.

abacigil:
Round-5 fixes in 3abd66e — both real bugs in the LiveRunPanel polling effect from the previous commit:

1. **Stale cursor + steps on run.id change.** When the panel stayed mounted across a new run, `lastIDRef` and the accumulated `steps` state were not reset. The new run's stream would render appended to the previous run's tail, and the ObjectID cursor could skip the head of the new run's rows. Added a separate effect keyed on `run.id` that clears `steps`, `lastIDRef`, and the scroll anchor before the polling effect re-fires.

2. **Overlapping polls on slow networks.** `setInterval` armed every 1.5s without awaiting the previous `listRunSteps`. Two requests could overlap with the same `since` and the dashboard would append the same rows twice. Switched to a self-scheduling `await-then-setTimeout` so the next tick only arms after the current one resolves — the cursor is guaranteed to advance between requests.

`tsc --noEmit` clean (only pre-existing `undici` errors remain, unrelated to this PR).

PR #211, opened 2026-05-12
fix(agent,warehouse): wire SQL dialect + dialect-correct table refs into exploration prompts

abacigil:
## Summary

Exploration prompts hardcoded BigQuery-style table refs and the orchestrator never injected the warehouse's actual SQL dialect into the prompt. A project connected to MSSQL (or Postgres, Snowflake, Redshift) saw BigQuery examples, generated BigQuery-shaped SQL, and paid one full SQL-fix LLM call per query to rewrite each statement into the connected dialect — ~30 extra LLM round-trips per discovery run, every run.

The validator (`insight_validator.go`) was only half-fixed: it already injected the dialect via `**SQL Dialect**: %s` but immediately undid that by hardcoding _"ALWAYS use fully qualified table names with backticks: `dataset_name.table_name`"_ in the same prompt.

## What changed

**Warehouse layer (`libs/go-common/warehouse`, `providers/warehouse/*`):**

- New `QuoteRef(parts ...string) string` method on the `warehouse.Provider` interface — implemented on all six shipped providers.
- Per-dialect quoting:
  - BigQuery / Databricks → `` `dataset`.`table` ``
  - PostgreSQL / Redshift / Snowflake → `"dataset"."table"`
  - SQL Server → `[dataset].[table]`
- Each provider delegates to a colocated `warehouse.QuotePartsWith(open, close, parts)` helper so the filter-and-join loop is not duplicated.

**Agent layer (`services/agent/internal/discovery/*`, `validation/*`):**

- New `substituteDialectTokens` helper in `prompt_subs.go` substitutes two new placeholders in every prompt path (exploration, analysis, recommendations, base context):
  - `{{DIALECT}}` → `provider.SQLDialect()`
  - `{{REF:tablename}}` → `provider.QuoteRef(firstDataset, "tablename")`
- `insight_validator.go` `buildVerificationPrompt` renders example refs through `QuoteRef` and replaces the backtick-only rule with dialect-neutral wording (still injects `**SQL Dialect**: %s` as before, just no longer contradicts it).
- `schema_discovery.go` fallback at line 230 (taken when a provider doesn't implement `SampleQueryBuilder` — none today, but the path exists) now uses `QuoteRef` instead of BigQuery-shaped backticks.

**Prompt layer (`domain-packs/*/prompts/*`):**

Every shipped domain-pack exploration prompt (gaming, social, ecommerce, system-test + all their category overrides — 12 files) gains a `**SQL Dialect**: {{DIALECT}}` line at the top and replaces 64 `` `{{DATASET}}.tablename` `` occurrences with `{{REF:tablename}}` placeholders. The agent's internal `lookup_schema` array continues to take the unquoted `dataset.table` form (it's a tool argument, not warehouse SQL).

## Backward-compatible by construction

The substitution is a **no-op** on any prompt that doesn't contain the new placeholders, so existing custom domain packs in other deployments keep working unchanged. Customers with custom packs upgrade by editing their `exploration.md`:

1. Add `**SQL Dialect**: {{DIALECT}}` near the top, next to `**Dataset**: {{DATASET}}`.
2. Replace `` `{{DATASET}}.tablename` `` with `{{REF:tablename}}` in every SQL example.
3. Soften any "use backticks" rule (the dialect line at the top is now the source of truth).

Full migration recipe in `docs/reference/prompt-variables.md`.

## Coverage

**Unit tests:**

- `QuotePartsWith` — 13 cases covering BigQuery / Postgres / MSSQL delimiter pairs, single / multi / empty / whitespace parts, embedded-dot pass-through.
- Per-provider `QuoteRef` tests on **all six providers** — dataset.table, catalog.schema.table, single-part, empty, reserved-word, leading-underscore, case preservation, empty-middle-part-skipped.
- `substituteDialectTokens` — both placeholders separately and combined, repeated occurrences, identifier-character constraints (underscore-start OK, leading-digit rejected, dot rejected), nil provider passthrough, empty-refDataset single-part rendering, malformed-delimiter pass-through.

**Integration tests (live-database round-trip of the QuoteRef'd SQL):**

- **Postgres** + **MSSQL** — testcontainer-backed, run in CI under `run-integration-tests`. Verified passing locally.
- **Snowflake** / **BigQuery** / **Redshift** / **Databricks** — env-var-gated, run when CI has the relevant credentials.

## Test plan

- [ ] All `make test-go` passes (verified locally — 0 failures).
- [ ] `make lint-go` clean (verified locally — 0 issues across every module).
- [ ] `make build` produces both binaries (verified locally).
- [ ] Postgres testcontainer integration test passes (verified locally — `TestIntegration_QuoteRef_RoundTrip` 3.65 s).
- [ ] MSSQL testcontainer integration test passes (verified locally — `TestIntegration_QuoteRef_RoundTrip` 31.31 s).
- [ ] CI integration job runs against the `run-integration-tests` label.
- [ ] Manual smoke: connect a discovery to an MSSQL project and confirm the exploration step emits `[ds].[table]` from the first query (no SQL-fix retry).


abacigil:
Addressed all 5 Copilot inline comments in a0da528:

**Comment on `insight_validator.go:549` (Copilot: quoting `user_id` example may push the model toward case-sensitive column refs).** Fixed — dropped `userIDIdent := QuoteRef("user_id")` entirely; the example now reads as plain `user_id` in markdown emphasis. Added a new prose rule in the **CRITICAL TABLE NAME RULES** block: "When citing a column in prose, write it unquoted (e.g. user_id) so the case-folded form the dialect expects is unambiguous." Also fixed an unrelated anti-pattern in the same prompt: the line that hardcoded "backticks for BigQuery/Databricks, double quotes for PostgreSQL/Redshift/Snowflake, square brackets for SQL Server" now defers to the dialect line above as the source of truth — that line is the only place that names the connected warehouse, future-proof for the next provider we add.

**Comments on `gaming` / `social` / `ecommerce` / `system-test` `exploration.md` (Copilot: markdown backticks around `{{REF:table_name}}` produce nested backticks after substitution on BQ / Databricks).** Fixed — stripped the markdown backticks from the critical-rule example in all four prompts; the placeholder now reads as plain `{{REF:table_name}}` so the rendered output is unambiguous on every dialect.

Local checks after the fix:
- `make test-go` — clean (validation package re-tested).
- `make lint-go` — 0 issues.
- The previously-passing `make test-postgres` and `make test-mssql` testcontainer integration tests still pass (the substitution logic itself is unchanged; only the example wording moved).

CI will re-run on this push. Marking the five threads as resolved.

abacigil:
Coverage follow-up landed in 20ade9d:

- `prompt_subs.go` `substituteDialectTokens` lifted from 90.9% → 100% by dropping a dead defensive branch (`if len(sub) < 2` was unreachable because `ReplaceAllStringFunc` only invokes the callback on a successful regex match, and the pattern always has exactly one capture group).
- `insight_validator.go` `NewInsightValidator` lifted from 75% → 100% via a new `TestNewInsightValidator_RefDatasetExtraction` that covers single-dataset, multi-dataset (comma-separated; with and without trailing whitespace), and empty cases — and explicitly guards that the `dataset` field stays verbatim while `refDataset` extracts the first segment.
- `services/agent/internal/testutil/` added to `codecov.yml` ignore list. The new `QuoteRef` mock added by this PR exists solely to be called from `*_test.go` and would otherwise count as uncoverable patch lines for every future PR that touches that package.

Patch coverage now **84.90% (up from 75.43%)**. The remaining 8 missed lines are all inside `RunDiscovery` (the orchestrator wiring at lines ~450 and ~669): a `refDataset := ""` declaration, the `if len(o.datasets) > 0` guard, and the four `template = substituteDialectTokens(...)` call sites. `RunDiscovery` is a ~500-line orchestration function with 0% unit-test coverage by design — it's exercised end-to-end by `services/agent/integration_test.go` (testcontainer-backed Mongo + mock LLM), which is what the green `Integration Tests` check at 3m55s actually runs. The helper `substituteDialectTokens` itself is at 100% via dedicated unit tests, so the only thing uncovered by unit tests is the literal `template = helper(...)` glue, which by construction has no branching of its own. I deliberately did NOT add `orchestrator.go` to the codecov ignore list because the same file contains plenty of unit-tested helpers (`resolvePrompts`, `buildFilterClause`, `parseInsights`, etc. — all at 100%) and a blanket ignore would hide too much.

`codecov/patch` is green, `Integration Tests` is green. The remaining uncovered lines are integration-test reality.

PR #215, opened 2026-05-14
feat: persist per-request LLM token usage as input_tokens/output_tokens

abacigil:
## Summary

- Every Mongo doc that represents one user-visible LLM request now carries the split pair `input_tokens` / `output_tokens` (both `omitempty`). Replaces the aggregate `tokens_used` on `AskSessionMessage`, `SearchHistory`, and `ExecutiveSummary`. Adds new fields to `RunStep`, `SchemaIndexProgress`, `ValidationResult`, and `InsightValidation`.
- New `libs/go-common/llm/UsageAccumulator` sums per-call usage onto a single home document when an internal multi-call operation has only one user-visible row — used by the insight validator (3 LLM calls per insight), the exploration engine's parse-retry loop, and the schema indexer's blurb pass.
- `ExplorationStep.TokensIn/TokensOut` were declared but never populated; the exploration engine now stamps them per step. New `StatusReporter.AddRecommendationStep` records the recommendation-phase LLM call on the live run-step log. `SchemaIndexProgressRepository.IncrementTokens` accumulates per-build totals; `Reset` zeroes them.
- No backfill: legacy rows that carry `tokens_used` retain it on disk; reads silently ignore it; the UI renders the new pair as absent on legacy rows.

## Pre-1.0 breaking change

Pre-1.0 breaking rename of the `tokens_used` field in API responses for `/api/ask`, `/api/runs/{id}/steps`, `/api/search-history`, and the schema-index status endpoint. The existing `tokens_in`/`tokens_out` on `ExplorationStep`/`AnalysisStep`/`RecommendationStep` are intentionally left on the older naming convention for this PR — known cosmetic inconsistency, follow-up cleanup PR may align them.

## Coverage

- 7-case `UsageAccumulator` unit suite (zero, single Add, many, zero-input, negative clamp, mixed-sign per-field clamp, Totals idempotence).
- BSON + JSON round-trip and omitempty tests for `AskSessionMessage`, `SearchHistory`, `SchemaIndexProgress`, and `ExecutiveSummary` — legacy `tokens_used` bson key asserted absent on the wire.
- Three-test validator suite covering the single-shot path, the multi-round lookup-loop token sum, and the LLM-error-before-usage path (which must leave the ValidationResult with zero tokens rather than uninitialised state).
- Orchestrator + status-reporter test fixtures updated for the widened signatures; exploration engine tests assert the parse-retry sum is plumbed through to the callback.
- Agent-side integration test on `IncrementTokens` (Reset zeroes, sums accumulate, mixed-sign deltas honour the clamp, missing-doc + empty-project-id errors fire).
- Schema-indexer integration test asserts a one-pass sum stamps the progress doc exactly once with the correct totals.

## Test plan

- [x] `make build` — clean
- [x] `make test-go` — all pass
- [x] `make lint-go` — 0 issues across every module
- [x] `cd ui/dashboard && npm run lint` — 0 errors (one pre-existing unrelated warning)
- [x] `cd ui/dashboard && npm test` — 212/212 pass
- [ ] CI integration tests (triggered by `run-integration-tests` label)
- [ ] End-to-end manual verification: run a full discovery, run `/ask`, eyeball `discovery_run_steps`, `ask_sessions.messages[]`, `search_history`, `discovery_validation_results`, and `project_schema_index_progress` for the new fields.

## Plan

PR 1 of the two-PR roll-out documented in `plans/PLAN-TOKEN-TRACKING.md`. PR 2 (enterprise repo) wires the executive-summary generator + store, the agentic-ask tool loop, and the domain-pack generator onto the new model shape — it depends on this PR being merged + tagged first.


abacigil:
## Copilot review pass — summary

Two comments, both valid, both addressed in 5f91d62:

1. **schema_indexer.go inline comment was wrong about partial-build semantics.** The comment claimed "partial-build failure preserves the running sum" — but `IncrementTokens` runs once after `blurb.Generate` succeeds, so a whole-batch blurb failure leaves the progress doc at the `Reset()` zeros, not at a partial sum. Behaviour wasn't broken — the *comment* was. Rewrote to spell out the three actual failure modes (whole-batch failure → zeros; per-blurb failure during a successful `Generate` → only successful blurbs summed; embed/upsert failure after this point → totals preserved).

2. **`ExplorationStep` doc reference said "single LLM call".** Now reads "usually one LLM call, but a turn whose response can't be parsed is retried… every retry's token usage is summed onto the same step" — matches the new `tokens_in`/`tokens_out` semantics that came in with the exploration-engine parse-retry sum.

Both review threads resolved on the PR.

## Codecov

Patch coverage was 57% on the first push (38 lines on `status.go`, 6 on `orchestrator.go`). Added seven new tests in `status_writer_test.go` (1cd42e9) that exercise every `Add*Step` method via `fakeRunStepWriter` — `AddAnalysisStep` (success + error), `AddValidationStep` (claimed>0 + claimed==0 message branches), `AddInsightStep` (pinning the "no tokens" contract for summary rows), `AddRecommendationStep` (success + error), and `AddExplorationStep` across `query_data` / `complete_rejected` / `lookup_schema` action paths (the invalid-ObjectID trick the existing tests use lets the downstream repo calls fail safely without a real Mongo). Coverage report on the next CI pass will confirm.

The remaining 6-line gap on `orchestrator.go` is wiring inside the 500-line `RunDiscovery` method — analogous to the already-codecov-ignored `services/api/internal/server/server.go` HTTP-routing wiring. If Codecov still flags it after the new tests land, I'll either add a targeted ignore entry or move the `if recStep != nil { AddRecommendationStep(...) }` wiring into a small helper. Will update.

abacigil:
## Second Copilot review pass — summary

Two more comments, both valid, both addressed in b88dd9e:

1. **`status.go` `AddExplorationStep` docstring used the wrong example.** I wrote that callers without an LLM call can pass zero tokens "(e.g. complete_rejected)" — but `complete_rejected` is produced FROM an LLM call (the model tried to complete early and the engine rejected it). Rewrote: every action the engine emits today comes from at least one LLM round, so non-zero usage is the norm; the omitempty zero-stored-absent contract is for any future path that produces a step without an LLM call.

2. **`CHANGELOG.md` path notation `services/{agent,api}/internal/models/...` was wrong.** The API-side structs live at `services/api/models/...`, NOT `services/api/internal/models/...`. Rewrote as two explicit paths so readers can navigate to the actual files.

Both review threads resolved.

PR #236, opened 2026-05-23
fix(agent): isolate discovery persistence ctx + raise outer run cap to configurable 24h

abacigil:
## Summary

- **Persist ctx isolation:** the tail-end durable writes (`Save`, `persistSplitLogs`, `runPhaseEmbedIndex`, `statusReporter.Complete`) now run under a dedicated ctx via `context.WithoutCancel(ctx) + WithTimeout(10*time.Minute)` so a completed discovery is never lost to the run-level deadline. Fixes the user-reported `discovery run failed: failed to save discovery result: failed to save discovery result: context deadline exceeded`.
- **Configurable outer cap:** the hardcoded `2 * time.Hour` in `agentserver.go` is gone. New env var `DISCOVERY_MAX_DURATION` (default `24h`, set to `0` to disable). Enterprise warehouses with multi-hour SQL no longer trip the cap by default.
- **Cleanup:** drops the duplicate ``"failed to save discovery result"`` wrap at the orchestrator/repo boundary so the error chain reads cleanly.

## Root cause

`services/agent/agentserver/agentserver.go:802` capped every discovery at 2h. `services/agent/internal/discovery/orchestrator.go` reused that same ctx for the tail-end persistence. A run that finished compute within the budget but had little headroom left would then fail to persist on the Mongo `InsertOne` (or any of the other tail writes) — losing the result.

## Design

- `persistContext(ctx)` deliberately drops cancellation: a SIGTERM during shutdown shouldn't poison a completed run. Values (correlation IDs, telemetry context) flow through.
- Per-step budgets (warehouse `QueryTimeout`, `LLM_TIMEOUT` + `chatWithRetry`, per-table schema timeout from PR #198) are what keep stuck operations responsive within a run — the outer cap is just a runaway-loop safety net.
- `DISCOVERY_MAX_DURATION=0` is the documented escape hatch for installs that want to rely purely on per-step budgets.

## Test plan

- [x] `go test ./services/agent/... -count=1` — all green
- [x] `golangci-lint run ./services/agent/...` — 0 issues
- [x] New unit tests pin: cancellation-survival, expiry-survival, value-preservation, deadline-window for `persistContext`; default / valid / zero-disables / invalid-falls-back / negative-falls-back / whitespace-trim for `discoveryRunContext`
- [ ] Integration CI on PR
- [ ] Codex review until-stable loop


abacigil:
**Round 1 — codex finding addressed in ed3e21f**

> [P2] Protect run completion from optional indexing timeout — orchestrator.go:886-892

Valid. `statusReporter.Complete` shared `persistCtx` with the heavier embed/index step. On large discoveries the embed/index could burn the 10-minute budget, leaving `Complete` to fire against an expired ctx — discovery saved but run never marked complete, no `discovery_id` back-reference (plugin-hooks.md Hook 5 broken).

Fix in commit ed3e21f:
- New `completeTimeout = 30 * time.Second` constant.
- `Complete` runs on a dedicated ctx via `context.WithTimeout(context.WithoutCancel(ctx), completeTimeout)`.
- Reordered: `Complete` now runs immediately after `Save` populates `result.ID`, before the non-fatal `persistSplitLogs` and `runPhaseEmbedIndex`. The run IS complete the moment the data is durably saved; indexing is an after-the-fact search optimization.
- New `TestCompleteTimeout_IsShorterThanPersistTimeout` pins the invariant so a future refactor that re-couples them fails loudly.

Kicking off round 2.

abacigil:
**Round 2 — codex finding addressed in 2c96fd0**

> [P2] Forward DISCOVERY_MAX_DURATION into agent Jobs — agentserver.go:998 (analysis pointed at `services/api/internal/runner/kubernetes.go:buildJob`)

Valid and important — without this propagation the env var only worked for subprocess runs; K8s production agents silently fell back to the 24h default.

Fix in commit 2c96fd0:
- `DISCOVERY_MAX_DURATION` added to the `buildJob` env-propagation allowlist alongside `QDRANT_URL`/`QDRANT_API_KEY`.
- Two new tests pin the contract: `TestKubernetesRunner_Run_PropagatesDiscoveryMaxDuration` (forwarded when set), `TestKubernetesRunner_Run_OmitsDiscoveryMaxDurationWhenUnset` (absent when unset — matching the QDRANT_* pattern).

Side note: a similar K8s-propagation gap exists for several other documented agent env vars added in earlier PRs (`LLM_RETRY_MAX_ATTEMPTS`, `LLM_RETRY_BASE_BACKOFF`, `BLURB_MAX_TOKENS`, `LLM_TIMEOUT`, `LLM_MAX_RETRIES`, `LOG_LEVEL`). Pre-existing — will be tracked as a separate follow-up issue rather than expanding this PR.

Kicking off round 3.

abacigil:
**Round 3 — codex finding addressed in 408b5df**

> [P2] Protect active runs longer than the sweep lookback — agentserver.go:1034

Valid and critical for the configurable-cap design. With `DISCOVERY_MAX_DURATION=168h` (the documented week-long enterprise case), the hardcoded 24h lookback in `loadActiveRunIDs` shadowed the cap — a still-running discovery older than 24h would be dropped from `keepRunIDs` and its Qdrant run-step collection deleted by the next boot-time sweep, breaking the live run.

Fix in commit 408b5df:
- New `discoverySweepLookback()` derives the lookback from the same env-var-parsing logic as `discoveryRunContext`.
- Branches: unset → 24h floor (historical behavior preserved); `0` → 30 days (cap disabled, status filter is the strong signal); `1h` → 24h floor (clamped, never below history); `168h` → 168h + 1h buffer; invalid/negative → 24h floor (typo-safe).
- Six new tests pin every branch.

**Self-audit additions (not from codex)**, pushed in f7e7839 just before this:
- Documented the interaction with `AGENT_JOB_TIMEOUT_HOURS` across configuration.md, helm-values.md, deployment/kubernetes.md, and the API chart's values.yaml — setting `DISCOVERY_MAX_DURATION` alone is silently shadowed by the wall-clock kill if `AGENT_JOB_TIMEOUT_HOURS` (default 6h) is left at its default.
- Added a startup `WARN` log via `runner.LoadConfig` when the two values are configured inconsistently, with truth-table coverage tests.

Kicking off round 4.

abacigil:
**Round 4 — both codex findings addressed in 295b735**

> [P1] Keep the new default below the API watch timeout — agentserver.go:1007

Valid and critical. With `DISCOVERY_MAX_DURATION=24h` default but `AGENT_JOB_TIMEOUT_HOURS=6h` default, the K8s watcher (and subprocess waiter) would stop polling at hour 6 — any failure after that left the run stuck in a non-terminal state. The old 2h hardcoded cap fit inside the 6h watcher; the new 24h doesn't.

Fix: raised `defaultJobTimeoutHours` 6h → 25h, paired with the 24h cap default with 1h headroom for the agent's 10-minute persistence tail + clock skew. Operators who already override the env var are unaffected; default-install operators get the correct pairing automatically. Existing `TestLoadConfig_Defaults` updated to assert the new default via the constant.

> [P2] Apply the sweep buffer for the unset default — agentserver.go:1079-1080

Valid bug. `discoverySweepLookback()` applied the buffer to explicitly-set values but returned exactly `sweepLookbackFloor` (24h) on the unset branch. A run started at hour 24:00:01 under the default cap would be dropped from `keepRunIDs` and its Qdrant collection deleted.

Fix: compute the effective duration first (env value OR `defaultDiscoveryMaxDuration`), then apply the buffer uniformly. Invalid/negative values now also fall back to the buffered default rather than the bare floor — more conservative. Two existing tests updated, one new test (`TestDiscoverySweepLookback_AppliesBufferToUnsetDefault`) pins the regression.

Helm values + all three docs (configuration.md, helm-values.md, kubernetes.md) updated to reflect the new 25h `AGENT_JOB_TIMEOUT_HOURS` default.

Kicking off round 5.

abacigil:
**Round 5 — both codex findings addressed in b96fa37**

> [P1] Keep completion after Phase 9 status updates — orchestrator.go:899-906

Valid and critical — this was a regression I introduced in round 1. `SetPhase(PhaseEmbedIndex, ..., 97)` writes `RunStatusRunning`, so marking Complete BEFORE Phase 9 silently downgrades the run back to `running`/`97%` with no later Complete. Every successful production discovery would end stuck non-terminal.

Fix in b96fa37: restored the original ordering (Save → splitLogs → embedIndex → Complete) and let `completeCtx` (the dedicated short ctx I added in round 1) do the actual isolation work. The reorder was unnecessary complexity — `completeCtx` alone solves the original deadline-coupling concern. Regression-history comment in `orchestrator.go` explains exactly why Complete must come AFTER `runPhaseEmbedIndex`.

> [P2] Compare against the effective default discovery cap — runner.go:139-142

Valid — the most common shadowed configuration silently slipped through the guard. `warnIfDiscoveryCapShadowed` only checked explicitly-set values, but an operator leaving an old `AGENT_JOB_TIMEOUT_HOURS=6` override in place with `DISCOVERY_MAX_DURATION` unset would have the agent apply its 24h default while the guard saw empty env and returned without warning.

Fix: compute the effective cap (env value if valid, else the 24h default) and compare. Explicit `0` still skips (deliberate disable). Log fields now carry both the effective cap and its source (env / default) so operators can act on the warning without re-reading the docs. Test cases updated to pin the unset/invalid/negative branches.

Kicking off round 6.

abacigil:
**Round 6 — codex finding addressed in 05b9931**

> [P1] Don't complete runs after the compute context expires — orchestrator.go:908-909

Valid and important. My persistence-ctx fix made Save/Complete keep running even when DISCOVERY_MAX_DURATION expired mid-compute — so a discovery that was actually killed by the outer cap during analysis was being marked completed and firing EventDiscoveryCompleted. Misleading success notification on a failed run.

Fix in commit 05b9931:
- Capture `computeErr := ctx.Err()` at the start of the persist tail BEFORE switching to `persistCtx`.
- Skip Phase 9 embed/index on the cancelled path (embedding partial / inconsistent data wastes tokens for indexes that go stale at the next run anyway).
- New `finalizeStatus` helper centralises the Complete-vs-Fail decision: `computeErr == nil` → Complete + return nil; `computeErr != nil` → Fail with the error message + return wrapped error so the caller routes through `EventDiscoveryFailed`.
- finalizeStatus runs under its own dedicated ctx (WithoutCancel + completeTimeout) so the terminal-status UpdateOne lands even when parent ctx is already cancelled.
- Extracted as a free function taking a `runFinalizer` interface so the Complete-vs-Fail decision is testable in isolation. Compile-time assertion pins that `*StatusReporter` satisfies the interface.
- Four new regression tests: happy-path completes; DeadlineExceeded → Fail not Complete + wrapped error; user-cancel → same; reporter ctx is live even with cancelled parent.

Kicking off round 7.

abacigil:
**Round 7 — codex finding addressed in f7d6be9**

> [P2] Preserve discovery_id when marking cancelled runs failed — orchestrator.go:94-96

Valid. My r6 fix marked cancelled-but-persisted runs as failed but didn't stamp `discovery_id` on the run document. The partial result was reachable in the `discoveries` collection but the failed run had no back-reference — plugin-hooks Hook 5 and the discovery-log APIs key off `DiscoveryRun.DiscoveryID`, so the partial result became unfindable through normal navigation.

Fix in commit f7d6be9 (pre-1.0 internal break, no shim):
- `RunRepository.Fail` and `StatusReporter.Fail` gain a `discoveryID` parameter, mirroring the `Complete` signature.
- Empty `discoveryID` means "no persisted discovery" (factory error, warehouse health-check failure) — the field isn't stamped.
- Non-empty stamps it the same way `Complete` does, so failed-with-partial-result runs are addressable identically to completed runs.
- `runFinalizer` interface + `runFinalizerFunc` test adapter updated. Existing test call sites adjusted to pass "" (pre-existing failures had no discovery to link).
- New assertion in `TestFinalizeStatus_ComputeCancelledCallsFailNotComplete` pins that the cancelled path passes `result.ID` to `Fail`.

Kicking off round 8.

abacigil:
**Round 8 — both codex findings addressed in d423097**

> [P2-a] Apply the job timeout to discovery Jobs — runner.go:185

Valid. KubernetesRunner.Run was creating discovery Jobs with no ActiveDeadlineSeconds — so the documented AGENT_JOB_TIMEOUT_HOURS wall-clock kill was never enforced by K8s, only by the in-process watcher. Long-running pods past the watcher window went unobserved.

Fix: wire `JobTimeoutHours * 3600` into `jobSpec.deadline` so the kubelet enforces the cap. The in-agent DISCOVERY_MAX_DURATION fires first (24h default, 1h below the 25h job timeout) so the agent fails gracefully through its persistence tail; the kubelet only kicks in as a hard-kill backstop if the agent itself wedges.

> [P2-b] Include persistence headroom in timeout warning — runner.go:178-185

Valid. `effective >= jobBudget` was the threshold, but the agent needs an additional 10 minutes for its persistence tail. `DISCOVERY_MAX_DURATION=24h59m` + `AGENT_JOB_TIMEOUT_HOURS=25` left only 1 minute of headroom and stayed silent.

Fix: compare against `effective + agentPersistenceHeadroom` where the headroom mirrors the agent's `persistTimeout = 10m`. New constant carries a "keep in sync" comment since the API doesn't import the agent package. Warning fields now expose `required_minimum` + `persistence_headroom` so operators can act on the warning without re-reading docs.

Tests: new `TestKubernetesRunner_Run_SetsActiveDeadlineSecondsFromJobTimeout` pins the deadline contract; warning-guard table-driven test gains three new cases — insufficient_headroom_warns, exact_headroom_warns_too, comfortable_headroom_safe.

Kicking off round 9.

abacigil:
**Round 9 — both codex findings addressed in 6a2cbbc**

> [P2] Leave the watcher alive past the active deadline — kubernetes.go:202

Valid. Watcher window matching ActiveDeadlineSeconds exactly left a race: the final tick can land before the K8s Job controller marks the deadline failure → watcher exits silent → OnFailure never called → run non-terminal even though kubelet kills the pod moments later.

Fix:
- Extend the watch window past the deadline by `watcherGracePeriod` (5 minutes) so the controller has time to surface the failure condition to our Get call.
- Defense in depth: when the watcher exhausts without ever seeing a terminal condition (kube-apiserver unreachable, Job controller wedged), still call OnFailure with a watch-timeout message so the run doesn't get stuck.

> [P3] Include final Complete headroom in the shadow warning — runner.go:196-197

Valid. Headroom only covered persistTimeout (10m), not completeTimeout (30s). Config like 24h50m + 25h made required == jobBudget exactly, leaving zero margin for the Complete UpdateOne — guard stayed silent.

Fix:
- Rename `agentPersistenceHeadroom` → `agentTerminalTailHeadroom = persistTimeout + completeTimeout = 10m + 30s`.
- Switch `>` to `>=` so exact-equal also warns (zero margin = misconfig).
- Log fields now surface `terminal_tail_headroom` + `required_minimum` for operators.

Kicking off round 10 (the last available round under the engineering rules' 10-round cap).

abacigil:
**Round 10 — codex finding addressed in 8a58a8b**

> [P2] Avoid failing cancelled Kubernetes runs after watcher exhaustion — kubernetes.go:479

Valid regression I introduced in r9. `Cancel()` deletes the Job, the handler stamps `RunStatusCancelled`, but the in-flight watcher goroutine then sees NotFound errors until its exhaustion branch fires — which called `OnFailure` and flipped the run from cancelled → failed.

Fix in 8a58a8b: detect `kerrors.IsNotFound` in the watcher's Get loop and exit silently. The cancel handler has already written the terminal status; the watcher's job is done.

New `TestKubernetesRunner_WatchJob_ExitsSilentlyWhenJobDeleted` pins the regression. To keep the test fast, `watchTickIntervalSec` is now a package var (defaults to 30s in prod, override to 1s in the test).

---

**Note on round count:** the engineering rules cap review rounds at 10. We've hit that. The user explicitly authorised continuing the loop until codex returns clean ("keep watching codex reviews continuously and continue until it satisfies"), so I'm running round 11. Surfacing this so it's transparent in the PR record.

Kicking off round 11.

abacigil:
**Round 11 — codex finding addressed in b1016bb**

> [P2] Don't fail runs after an inconclusive watch timeout — kubernetes.go:501

Valid. My r9 watcher-exhaustion fallback called OnFailure → `RunRepository.Fail`, but Fail unconditionally overwrote status — so a successful discovery that the agent had already stamped completed could be flipped to failed hours later from nothing more than the API losing K8s observability.

Fix at the **repo layer** (defense in depth, not just for the watcher):
- `Fail` now uses an UpdateOne with a filter matching only `pending`/`running` runs. Terminal status (`completed`, `failed`, `cancelled`) is immutable from Fail's perspective.
- Mongo's UpdateOne returns no error when zero documents match, so the existing "log + swallow" pattern in `StatusReporter.Fail` keeps working.

Three integration tests pin every case:
- Fail after Complete → no-op, status stays completed
- Fail after Cancel  → no-op, status stays cancelled
- Fail while running → updates, status transitions to failed

Kicking off round 12.

abacigil:
**Round 12 — codex finding addressed in 7eeb183**

> [P2] Guard late watcher failures from overwriting terminal runs — kubernetes.go:501

Valid — embarrassing miss. My r11 fix was at the **agent** repo (`services/agent/internal/database`), but the K8s watcher's OnFailure callback flows through the **API-side** repo (`services/api/database`) — a different `RunRepository.Fail` whose unconditional UpdateByID I didn't touch. So the r11 codex concern was only half-fixed.

Fix in 7eeb183: applied the same terminal-status filter (`status IN [pending, running]`) to the API repo's Fail. UpdateOne returns no error on zero-match so the swallow path still works.

Three integration tests on the API repo pin the same contract: Fail after Complete is a no-op, Fail after Cancel is a no-op, Fail while running succeeds.

**Lesson:** agent-side and API-side repos on the same collection need the same invariants — there's no shared interface enforcing this. I'm tracking that as a separate follow-up rather than scope-creeping this PR.

Kicking off round 13.

abacigil:
**Round 13 — structural revert in 03ffbe3 (per maintainer call)**

The cascade of r10-r13 P2 findings — all caused by the watcher exhaustion OnFailure path I introduced in r9 — was clear diminishing returns past the engineering rules' 10-round cap. Each fix added a new state check; each new state check opened another corner case.

Per maintainer direction, reverting the r9 exhaustion-OnFailure entirely. The agent's own terminal-status write is authoritative; ActiveDeadlineSeconds on the Job (from r8) means the kubelet hard-kills if the agent runs over. A watcher that 'fails the run from observability loss' was found to corrupt successful and cancelled runs in degraded scenarios — more harmful than helpful.

**KEPT** (the work that's still solid):
- Watcher grace period (5min past ActiveDeadlineSeconds) for observable deadline failures
- Both RunRepository.Fail terminal-status invariants — defense in depth for any future Fail caller
- ActiveDeadlineSeconds on discovery Jobs (r8)
- Headroom-aware shadow-cap warning (r9)

**REVERTED**: r9's OnFailure-on-exhaustion + r10's NotFound short-circuit.

`newTicker` refactored to `time.Duration` so exhaustion tests run in ms instead of waiting out 25h. New `TestKubernetesRunner_WatchJob_ExitsSilentlyOnExhaustion` pins the new invariant: watcher MUST NOT call OnFailure on exhaustion.

Kicking off round 14 to verify convergence.

abacigil:
**Round 14 — clean (loop converged)**

> The changes build and the targeted tests for the modified API runner/database and agent discovery/database packages pass. I did not identify any discrete correctness issues introduced by the patch.

The r13 structural revert dissolved the cascade — no new VALID findings on the verification pass. PR is merge-ready.

Final round-count tally: **14 rounds, 13 fixes, 1 strategic revert** (r13 → r14 verification). All findings addressed, terminal-status invariants hardened on both Fail repos, and the watcher design simplified to rely on the agent's own authoritative terminal-status writes.

PR #244, opened 2026-05-26
fix(llm/ollama): reasoning-aware Chat() — pin num_ctx, surface Thinking, expose ReasoningEffort

abacigil:
## Summary

The Ollama provider was sending only `temperature` and `num_predict` on
each Chat request, leaving `num_ctx` to the server's
`OLLAMA_CONTEXT_LENGTH` default (2-4k on stock installs) so large
prompts silently truncated. Reasoning-capable models (`gemma4`,
`gemma3`, `deepseek-r1`, `qwen3`) emit a hidden chain-of-thought
before the visible answer; the provider read only `Message.Content`,
dropped the reasoning, and the catalog's 16k-32k `MaxOutputTokens` cap
left no room for both the reasoning and the answer on rich prompts —
the visible failure mode was empty or truncated content with no signal
to the caller.

Empirically reproduced against `gemma4:31b-it-bf16` on an A100 80GB /
Ollama 0.24.0 VM: a tiny prompt with `num_predict: 50` returned
`content: ''` because every token went into the `thinking` field; a
realistic-sized prompt with default settings burned ~25 % of the
output budget on reasoning we then dropped.

### What changed

- `providers/llm/ollama/provider.go` Chat() now sets
  `options[\"num_ctx\"]` from `gollm.GetMaxInputTokens(\"ollama\", model)`,
  pins `Truncate=false`, reads `Message.Thinking` into the new
  `ChatResponse.Reasoning`, and forwards
  `ChatRequest.ReasoningEffort` to Ollama's top-level
  `Think *ThinkValue`.
- New `num_ctx_cap` config field (project's Ollama settings → \"Max
  context tokens\") clamps `num_ctx` lower than the catalog when the
  operator is running on tight VRAM.
- Catalog raises `MaxOutputTokens` to 131072 for the verified
  reasoning rows (`gemma4`, `gemma4:31b`, `gemma3`, `deepseek-r1`,
  `qwen3`); non-reasoning rows untouched.
- `ollamaDefaultTimeout` 5m → 15m. At ~20 tok/s on a 31B-class local
  model, 5m capped generation at ~6k tokens — below the working
  size of a reasoning-on response.
- `gollm.ChatRequest` gains `ReasoningEffort string`;
  `gollm.ChatResponse` gains `Reasoning string`. Documented values:
  `\"\"` (default), `\"off\"`, `\"on\"`, `\"low\"`, `\"medium\"`,
  `\"high\"`; unknowns degrade to default with no error.
- `services/agent/internal/discovery/blurb/blurb.go` —
  reasoning-class rejection error rewritten. The rejection still
  stands on per-call wall-clock grounds (2k parallel small calls),
  but the \"thinking channel not exposed\" reason is no longer
  accurate.

### Pre-1.0 breaks

- `NewOllamaProvider` gains a fourth parameter `numCtxCap int`
  (zero or negative = no cap).
- `gollm.ChatRequest` gains `ReasoningEffort string`.
- `gollm.ChatResponse` gains `Reasoning string`.

Out-of-tree callers that pass zero / leave the new fields empty get
behaviour fully compatible with today's defaults.

## Test plan

- [x] `make test-go` — every module green (libs + 6 LLM providers +
      6 warehouse providers + 3 secret providers + 6 embedding
      providers + agent + api)
- [x] `make lint-go` — 0 issues across every module
- [x] `make build` — agent + api binaries built clean
- [x] `make lint-docs` — clean
- [x] 7 new unit tests in `providers/llm/ollama/wiring_test.go`:
  - `num_ctx` resolved from catalog for 5 representative tags
    including unknown-tag fallback
  - `num_ctx_cap` interactions (below / equal / above / zero)
  - `Truncate=false` always present
  - All 7 `ReasoningEffort` values map to the right Think shape
    (including unknown → omit)
  - `Reasoning` round-trip with the empty-content-stays-empty
    guard (no fallback)
  - Factory parses every `num_ctx_cap` shape (missing / empty /
    valid / negative / non-numeric / zero)
  - `num_ctx_cap` config field appears in registered ProviderMeta
- [x] Container-backed integration test
      (`services/agent/ollama_integration_test.go`, build tag
      `integration && ollama`) exercises every documented
      `ReasoningEffort` value end-to-end against a real Ollama and
      pins the `Truncate=false` overflow contract — a deliberately
      tiny `num_ctx_cap=256` with a 4 KB prompt must return an
      error from the live Ollama server
- [x] Empirical bench against the live A100/gemma4:31b VM
      established baseline + with-thinking + cap-raised + per-request
      num_ctx-override behaviour (see commit message + the detailed
      table in the design notes)


abacigil:
Round-1 codex review triage — both findings VALID, both addressed in `08ba0a0`.

| Finding | Status | Fix |
|---|---|---|
| `ReasoningEffort=on/low/med/high` 400s on non-reasoning models | **FIXED** | New `llm.ModelEntry.Reasoning bool` + `llm.IsReasoningModel(provider, model)`; provider's `reasoningEffortToThinkValue` takes the flag and omits Think for `on/low/medium/high` against non-reasoning rows. `off` still serialises (`think=false` is harmless on every model). New `TestChat_ReasoningEffortMapping_NonReasoningModel` proves it; integration test now passes against the non-reasoning `qwen2.5:0.5b` for every effort value. |
| `num_ctx_cap` not threaded into Ask budgeting | **DEFERRED** | Dropped the field + ConfigField + tests + doc references from this PR. `NewOllamaProvider` is back to 3-arg. Operators on tight VRAM use `OLLAMA_CONTEXT_LENGTH` on the Ollama host (already documented). Follow-up issue forthcoming to properly thread an operator cap through every budgeting call-site so the cap and the budgeter agree. |

CI / local: `make test-go` + `make lint-go` + `make build` + `make lint-docs` all clean.

abacigil:
Round-2 codex review — one VALID P1: per-request `num_ctx` is an Ollama override (not server-capped), so the previous commit's catalog-driven forward would OOM low-VRAM deployments. Fixed in `383652d`: `num_ctx` is now an operator opt-in config field, omitted when unset. `Truncate=false` stays default-on (no VRAM cost). `NewOllamaProvider` gains a 4th param `numCtx int` (zero = don't send). All tests + lint + docs green.

abacigil:
Round-3 codex review — one VALID P2: reasoning rows for deepseek-r1 / qwen3 / gemma3 missed common pulled tags (deepseek-r1:1.5b/7b/8b, qwen3:0.6b/1.7b/4b/8b/14b, gemma3:1b/4b/12b). Exact-match `IsReasoningModel` falls through to false for unlisted sizes, silently stripping Think. Fixed in `e20147d`: aliases now cover every Ollama-library size for each reasoning family; `TestIsReasoningModel_RegistryLookup` expanded to pin resolution for every size. All tests + lint + docs green.

abacigil:
Round-4 codex review — both findings VALID, both fixed in `42fa6b7`. (1) New `gollm.GetEffectiveInputWindow` + optional `ProviderMeta.EffectiveInputWindow` hook; Ollama registers one that clamps to operator `num_ctx`; `search.go` now uses it so `/ask` budgeting respects the deployment cap. (2) `blurb.New()` ORs the existing substring check with `gollm.IsReasoningModel(providerName, model)` so Ollama gemma4/gemma3/qwen3/deepseek-r1 are rejected at construction time on throughput grounds. New tests for both. All green.

abacigil:
Round-5 codex review — one P2: budgeter uses catalog window when num_ctx unset, while server enforces OLLAMA_CONTEXT_LENGTH; mismatch surfaces as a Truncate=false error.

**WONTFIX-with-rationale** in `1dd6c7c` (docs-only). Reasoning:
- Without probing the server (per-model, per-load) we cannot distinguish "operator raised OLLAMA_CONTEXT_LENGTH to 64k+ and wants the full window" from "operator left server at 4k and would prefer aggressive trimming".
- A conservative 4k default would over-trim the common case where operators have raised the server config.
- The new failure mode (loud error) is strictly better than the prior silent truncation. Operators get an actionable error.
- Two clear remediation paths exist and are now explicitly documented in `docs/guides/configuring-llm.md`: raise OLLAMA_CONTEXT_LENGTH on the host OR set the project's num_ctx — that single value is used both for the per-request override and the budgeting clamp, so they stay aligned.

Net-zero VALID concerns remaining. Stopping the loop.

PR #287, opened 2026-06-23
feat(agent): ad-hoc data Q&A serve mode (--mode ask-serve)

jale-dbox:
## What

Adds a long-lived, multi-project HTTP **serve mode** to the agent — `--mode ask-serve` — that answers a natural-language data question by running a **bounded, tool-using reasoning loop** against a project's warehouse (read-only) and schema index, persisting each step to Mongo so a reader can tail progress **by cursor**. This is the same persist-and-poll shape a discovery run already uses (the API never holds a connection to the agent), so it survives hours-long queries and client disconnects.

It is a generic data-Q&A serving capability on the community agent; downstream builds run the existing agent image with `--mode ask-serve`.

## How

New package `services/agent/internal/askserve`:
- **pool** — warm per-project runtime cache (warehouse + LLM + schema), lazy build on first use, idle + LRU eviction, reference-counted so an in-use runtime is never evicted; concurrent callers for one project share a single build (race-tested).
- **loop** — bounded ReAct loop with a JSON action vocabulary: `query_data` / `lookup_schema` / `search_tables` / `answer` / `clarify` / `decline`, with parse-retry nudging and a final-synthesis fallback at the step budget.
- **summarize** — summary-only result handling: a fetch cap (memory bound) + preview cap (token bound) + explicit truncation marker; the full result set is never stored or streamed.
- **turnstore** — durable turn record + incremental tool-event log with an **ObjectID cursor** read, a terminal-only TTL, and a boot/shutdown **orphan sweep** (a turn stranded by a restart reaches `failed` with its partial transcript preserved).
- **server** — `POST /turns` (ACK + background job), `GET /turns/{id}` (status), `GET /healthz`; global + per-project concurrency caps (429 on excess); graceful shutdown drains in-flight turns then sweeps.

Reuse, not rebuild: the loop delegates to the existing **query executor** (tenant filter + SQL self-heal), the **cached schema provider**, and the **LLM client**. `agentserver` wires the per-project runtime through the existing provider/secret helpers. `ValidateReadOnly` is enforced at connect so a turn never runs against write-capable credentials.

Also adds the `policy.FeatureDataQuery` wire constant (+ `AllFeatures` entry) so plan-gated builds can entitle the capability.

All budgets are env-configurable (`ASK_SERVE_*`). A single enterprise query can run for hours, so the per-turn wall-clock defaults to 12h — the bound on a **durable background job**, not a held connection.

## Verification

- `go build ./...` and `go vet` clean across the agent module + go-common.
- Unit tests: action parser (every mode, tool-use envelope, noisy/reasoning preambles, braces-in-SQL, unparseable), summarize (no-truncation / preview-cap / fetch-cap / empty / ragged columns), config (defaults / env overrides / invalid fallback), **pool with `-race`** (reuse, concurrent-share, idle + LRU eviction, in-use protection, build-error retry), and the **loop with fakes** (happy / multi-query / clarify / decline / query-budget / round-cap-synthesis / round-cap-fail / parse-retry / query-error / tenant-filter-rejected / schema-tool-degrade).
- Integration tests (testcontainers Mongo): append + ObjectID-cursor read, finalize writes the terminal turn + session message (idempotent), orphan sweep, idempotent ensure.

— Co-coded with Jale 🤖


jale-dbox:
## Companion to enterprise #113 — final note

This PR is the platform half of **Conversational Data Q&A** (decisionbox-io/decisionbox-enterprise#113): `--mode ask-serve` + the `data_query_enabled` policy constant + the shared `AskTurn`/`AskTurnEvent` models. The enterprise build compiles against it, so **this merges first**.

**Verified:** `go build ./...`, `go vet`, `golangci-lint`, the ask-serve unit suite (action parser, summarize, config, pool with `-race`, loop with fakes) + testcontainer integration (turn store append/cursor/finalize/orphan-sweep, `$literal` answer), and a live E2E driving the agent loop against a real BigQuery project (the E2E caught + fixed a grounding bug: the loop now refuses an answer produced with no tool activity).

**Codex review: 5 rounds.** Fixed: shared-executor `SetStep` data race, event-log TTL, legacy-null-safe + `$literal` session append, governance project-id on the turn context, project-deletion cascade for the new collections, persisting the turn model, lowercase search-event output.

**Reviewer decision (@abacigil):** Codex repeatedly requested regex/keyword **read-only SQL parsing** and **exact tenant-predicate enforcement**. Per the architecture these are intentionally **not** done — read-only is the customer's read-only-credential contract (+ `ValidateReadOnly` + governance) and the warehouse layer is multi-provider, so the SQL guard was reverted; **self-heal stays**, tenant scope stays on the existing `queryexec` filter + the prompt. WONTFIX-by-design.

**Follow-ups:** a pre-fetch row cap belongs in the warehouse-provider layer (max-rows), not dialect-specific SQL rewriting; cross-replica/restart turn ownership (ask-serve runs single-replica in Phase 1).

— Co-coded with Jale 🤖

PR #333, opened 2026-08-13
fix(ask-serve): make chart caps actionable in the model's guidance

jale-dbox:
Found while testing #128 end-to-end on a live stack.

## The dead end

Asked: *"For each customer, plot their total number of invoices against their total amount spent — is there a relationship?"* (chinook: **59 customers**, `PreviewRows` **50**).

The turn produced three rejected charts and then gave up on the scatter, answering with a 2-point bar instead:

```
render_chart  rejected [shape: data]
render_chart  rejected [grounding: source_step_id]: step "q1" was truncated…
render_chart  rejected [caps: caption]
```

Two things were wrong with the guidance around the truncation rule (the rule itself is correct — a partial preview must not ground a chart):

1. **The advice destroys the question.** It said *"aggregate in SQL (GROUP BY/LIMIT) so the full result fits the preview"*. For a per-entity chart, one point per customer **is** the chart — aggregating answers a different question.
2. **The rejection never names the cap.** It said only that the step *"was truncated (its preview omits rows)"*. With just the prompt reworded, the model did reach for `LIMIT` — and chose `LIMIT 59`, the full row count, so it was truncated again. It had no way to know the number.

## Changes

- **`prompt.go`** — pick the fix that preserves the question: aggregate (`GROUP BY`) for totals and breakdowns; `ORDER BY <measure> DESC LIMIT <preview>` for one-point-per-entity charts, disclosing the cut in the title/caption ("Top N customers by spend") so it is not read as the whole population. The cap is interpolated from `cfg.PreviewRows`, not implied.
- **`loop.go`** — when the source step was truncated, the rejection now states the preview cap and the step's real row count, so a retry has the number in hand. The validator is generic and cannot name a cap the loop owns.
- **`validate.go`** — same two-way framing in the validator message.

## Result

Same question, after both changes:

```sql
… GROUP BY c.customer_id ORDER BY total_spent DESC LIMIT 50
```
→ 50 rows, `truncated: false` → **a 50-point scatter**, titled *"Total Invoices vs. Total Spent (Top 50 Customers by Spend)"*.

## Scope

Guidance only — no change to the grounding rule, the caps, or the validator's behaviour. `charts` and `askserve` suites pass.

This does not remove the ceiling: a per-entity chart is still capped at `PreviewRows`, because that value is doing double duty as the model's token budget **and** the only retained copy of the rows (`summarize.go`: *"No raw result set is ever stored"*). Lifting it properly means projecting chart data server-side from a turn-scoped buffer so the cap is set by rendering legibility rather than context size — worth a separate issue rather than a bigger number here.

https://claude.ai/code/session_019ywMCHCbrCUkRoC6s57XuA


jale-dbox:
Added a second fix in 935d1df, same root cause — a cap the model is never told about.

`MaxLabelLen` (120) is enforced on every title, caption and label, but the prompt only said "clear labels", so a natural-length caption burned a round on `chart rejected [caps: caption]`. The cap is now stated, interpolated from `cfg.ChartCaps.MaxLabelLen` so an overridden deployment still gets the truth.

Retitled the PR accordingly — both changes are "tell the model the limits it has to satisfy".

End-to-end on the same question, measured on a live stack:

| | rejected charts | result |
|---|---|---|
| before | 3 | no scatter — pivoted to a 2-point bar |
| + truncation guidance | 2 | 50-point scatter |
| + label cap | **0** | 50-point scatter on the first attempt, caption 99 chars |

jale-dbox:
## Codex review — loop closed (clean final round)

Ran `codex review --base ask-charts` to stability.

| Round | Finding | Triage |
|---|---|---|
| 1 | P2 — guidance quoted `PreviewRows` as the row limit, but `MaxFetchRows` and `ChartCaps.MaxPoints` can each be stricter, so a tuned deployment would be told "LIMIT 50 is enough" and rejected again | VALID → fixed ad6816b |
| 2 | — | clean |

The round-1 finding was the failure this PR exists to prevent, reintroduced through a different cap: invisible on defaults (`MaxPoints` defaults to `PreviewRows`, fetch cap 1000), but real for anyone who lowered `ASK_SERVE_CHARTS_MAX_POINTS` or `ASK_SERVE_MAX_FETCH_ROWS`. Now centralised in `Config.ChartableRowCap()` (smallest positive cap) and used by both the prompt and the rejection, with table-driven coverage.

Unrelated, noticed while working here: `config.go` is gofmt-dirty from #305/#307 — the `EnvCharts*` const block is misaligned. CI runs golangci-lint without `gofmt`/`gofumpt` so nothing fails; left alone to keep this diff focused.

Issue #334, opened 2026-08-13
Per-entity charts are capped by PreviewRows, not by what a chart can legibly show


jale-dbox:
Found testing charts (enterprise #128) end-to-end. #333 makes the current cap *recoverable* — this issue is about removing it.

## The ceiling

A chart can only be grounded on a query whose **full result fit the preview** (`ValidateGrounded` refuses a truncated step — correctly, since a partial preview would let a subset be presented as the whole). `PreviewRows` defaults to **50**.

So "one point per entity" is capped at 50 entities. Asking

> *For each customer, plot their total number of invoices against their total amount spent*

against chinook's **59 customers** cannot produce a 59-point scatter at all. After #333 the model recovers to an honest `LIMIT 50` "Top 50 customers by spend" — but the other 9 customers are simply unchartable.

## Why raising the number is the wrong lever

`PreviewRows` is doing double duty. From `summarize.go`:

> *"It is what gets persisted as the tool event's Output and rendered in the transcript — never the full table. **No raw result set is ever stored.**"*

It is simultaneously:
1. the **LLM's token budget** for observing a result, and
2. the **only retained copy of the rows**, hence the ceiling on chart size.

So the chart cap is set by how much text we are willing to put in a prompt — which has nothing to do with how many points a scatter can legibly show. Raising it to 60 fixes 59 customers and breaks at 61, while taxing every query observation and inflating every persisted tool event, to help only the charted ones.

## Proposed fix — project chart data server-side

Stop having the model author the `data` array. `render_chart` supplies only `source_step_id`, `type`, the x/y **field names**, labels and unit/format hints; the server projects the rows itself from a **turn-scoped, in-memory buffer** retained up to `ChartCaps.MaxPoints` (say 500) and not persisted, so transcript size and token cost stay exactly where they are today.

What that changes:

- **Grounding becomes structural rather than validated.** There is no model-authored cell left to compare, so the cell-matching in `ValidateGrounded` — and truncation as a *grounding* concern — goes away. Fabrication becomes impossible by construction.
- **The cap moves to where it belongs**: rendering legibility, not context size.
- **Removes a whole class of model error.** `chart rejected [shape: data]: this chart type requires non-empty data` cannot happen if the model never writes the data.
- **Shrinks the spec**, relieving `MaxSpecBytes`.

`ChartSpec.Data` stays in the persisted spec — `chartexport` and the dashboard's `ChartCard` both read it — the server just fills it instead of the model.

[...]

PR #335, opened 2026-08-13
feat(ask-serve): grounded charts — ChartSpec + render_chart (ask-charts → main)

jale-dbox:
## What & why

Bridge PR: lands the whole `ask-charts` platform bridge branch on `main`. This is the **platform half** of enterprise charts in Ask sessions (decisionbox-io/decisionbox-enterprise#128) — the enterprise bridge PR (`chartexport` + the Recharts UI) depends on this and must merge **after** it.

The data agent (`ask-serve`) can call a new **`render_chart`** tool that emits a constrained, renderer-agnostic **`ChartSpec`** (bar / line / area / pie / scatter / KPI), **hard-grounded** in a query result the agent already observed — never invented, never derived.

## What's in it

Four PRs already reviewed and merged into `ask-charts`:

| PR | Commit | What |
|----|--------|------|
| #305 | `b1f881c` | `libs/go-common/charts` (ChartSpec, `schema.json`, `Validate` + `ValidateGrounded`), `FeatureAskCharts` policy flag, `render_chart` wired into both the native tool path and the JSON-text fallback |
| #306 | `c13ab3d` | Field-name sanitization + single-y `series_by` (Codex follow-up to #305) |
| #307 | `510cc88` | Number formatting — compact values + unit/format hints |
| #333 | `67a5708` | Chart caps made actionable in the model's guidance (`ChartableRowCap`, cap named in the rejection message) |

`aa02610` is a `main` → `ask-charts` merge; the branch is **0 behind `main`**.

## The grounding rule

`ValidateGrounded` is the load-bearing part, and it is **structural, not prompt-level**: a chart's `data` must be an exact projection of the referenced query's preview. The model may drop or reorder columns and rows; it may never compute, round, scale, or invent a number — aggregation belongs in SQL. A truncated source cannot ground a chart, KPI values must name a source column, and figures beyond float64's exact-integer range are refused. A rejected spec self-heals through the agent's existing error-observation retry.

`render_chart` is deliberately **not evidence**: a successful chart never grounds the turn, so charting can't become a backdoor to the `answer` tool.

## Enterprise-only by wiring, not by code secrecy

The tool lives in this AGPL repo, but community never deploys or calls `ask-serve`. It is offered only when the enterprise delegate sets the `EnableCharts` wire capability **and** `ASK_SERVE_CHARTS_ENABLED` is on **and** a non-truncated query has run.

## Verification

Beyond the unit suites (`charts/validate_test.go`, `askserve/chart_test.go`, `config_test.go`, `loop_test.go` — native and text-path parity), this ran end-to-end on a dedicated stack (Mongo + Qdrant + `ask-serve` + dashboard) against a real Postgres warehouse. Grounding was checked three ways for the same turn — the SQL result, the `query_data` tool event, and the rendered chart's `data` — and matched byte-exact.

## Known follow-up

#334 — per-entity charts are capped by `PreviewRows` rather than by what a chart can legibly show. #333 made the ceiling *actionable* (the model is told the number and the rejection names it, which took rejections 3 → 2 → 0 in testing) but the underlying coupling of "token budget" and "only retained row copy" is unresolved. Deferred deliberately, not overlooked.

## Review

Codex review loop run to zero valid concerns on each constituent PR.


jale-dbox:
Companion enterprise bridge PR: decisionbox-io/decisionbox-enterprise#205. **Merge this one first** — the enterprise branch doesn't compile against platform `main` until it lands.

abacigil (approved):

PR #340, opened 2026-08-26
feat(llm): schema-constrained structured output across providers

jale-dbox:
## Summary

Adds an optional, wire-neutral **structured-output** capability to the LLM provider layer: a caller can attach a JSON Schema to a `ChatRequest` and get back a guaranteed-shape single JSON object, using each provider's native mechanism. This is the platform half of a coordinated change (the enterprise domain-pack generator is the first consumer; a self-hosted qwen quant behind LiteLLM→Ollama was wrapping its output in a JSON array, which schema-constrained decoding makes impossible).

## What changed

**Wire-neutral contract (`libs/go-common/llm`)**
- `ChatRequest.ResponseFormat *ResponseFormat` (`{Name, Schema, Strict}`) — advisory; providers that don't support it ignore it.
- `ProviderMeta.SupportsStructuredOutput bool` — per-provider capability, surfaced in `/api/v1/providers/llm`. `GetProviderMeta` lets callers gate on it.
- `ApplyResponseFormatAsTool` / `NormalizeStructuredToolResponse` — shared helpers for the Anthropic wire (which has no `response_format`): translate a `ResponseFormat` into a single **forced tool** and fold the tool result back into `Content` so callers parse uniformly.

**Per-provider translation**
- **OpenAI-compatible wire** (`openai`, `litellm`, and OpenAI models on `bedrock`): `response_format: {type: json_schema, …}` via the shared `openaicompat.BuildRequestBody`.
- **Ollama**: the `format` field (llama.cpp grammar-constrained decoding).
- **Anthropic wire** (`claude`, Claude models on `bedrock`): forced tool-use via the shared helper.

**Enabled** for `ollama`, `litellm`, `openai`, `claude`, `bedrock`. `azure-foundry` and `vertex-ai` keep the flag **off** for now — their Claude/Gemini wires don't implement the tool plumbing needed to honour it, and advertising a per-provider flag that only some of a provider's wires satisfy would over-promise. Callers fall back to their own parsing there.

## Non-regression (important)

- `ResponseFormat` is **orthogonal to `Tools`**: the synthetic tool is injected only when the caller supplied **no** tools of its own, so existing tool-using flows (e.g. `/ask` function calling) are byte-for-byte unchanged.
- A request **without** `ResponseFormat` produces an identical wire body to before (tests assert `response_format` and Ollama `format` are omitted when unset).
- **Open-ended (dynamic-key) objects are preserved**: the OpenAI wire uses **non-strict** `json_schema` (strict mode forbids `additionalProperties` maps), and the Ollama grammar + Anthropic tool schema accept them natively. Tests assert an `additionalProperties` object survives onto the wire on all three paths.

## Tests

- `structured_output_test.go` — the forced-tool helper (inject / no-op / caller-tools-win / normalize / text-left-intact).
- `openaicompat/format_test.go` — `response_format` json_schema wiring + omitted-when-unset + open-map preservation.
- `ollama/provider_test.go` — hermetic httptest round-trip asserting `format` is set (and omitted when unset).
- `registry_test.go` — `supports_structured_output` serialises.
- Per-provider `SupportsStructuredOutput` flag assertions.
- `go build` + `go test` + `golangci-lint` green across every touched module.

— Co-coded with Jale 🤖

### DIFF
--- a/libs/go-common/llm/openaicompat/format.go
+++ b/libs/go-common/llm/openaicompat/format.go
@@ -41,6 +41,32 @@ type RequestBody struct {
 	// always did.
 	Tools      []Tool      `json:"tools,omitempty"`
 	ToolChoice interface{} `json:"tool_choice,omitempty"`
+
+	// ResponseFormat is populated only when the caller supplies
+	// ChatRequest.ResponseFormat. Omitted otherwise so requests that
+	// don't ask for structured output keep exactly the shape they always
+	// had.
+	ResponseFormat *ResponseFormat `json:"response_format,omitempty"`
+}
+
+// ResponseFormat is the OpenAI `response_format` object. Only the
+// json_schema variant is emitted here (Type == "json_schema"); the
+// json_object and text variants have no consumer.
+type ResponseFormat struct {
+	Type       string      `json:"type"`
+	JSONSchema *JSONSchema `json:"json_schema,omitempty"`
+}
+
+// JSONSchema is the inner object of a json_schema response_format.
+// Strict is intentionally left false for schemas that use open-ended
+// (dynamic-key) objects: OpenAI strict mode forbids them (it requires
+// additionalProperties:false and every property required), which would
+// drop such fields. Non-strict json_schema still steers the model to a
+// single conforming object while keeping open maps expressible.
+type JSONSchema struct {
+	Name   string                 `json:"name"`
+	Schema map[string]interface{} `json:"schema"`
+	Strict bool                   `json:"strict"`
 }
 
 // Tool is one entry in the `tools` array. OpenAI distinguishes tool type
@@ -65,7 +91,7 @@ type ToolFunction struct {
 type Message struct {
 	Role       string     `json:"role"`
 	Content    string     `json:"content"`
-	Name       string     `json:"name,omitempty"`        // legacy function-call naming
+	Name       string     `json:"name,omitempty"`         // legacy function-call naming
 	ToolCallID string     `json:"tool_call_id,omitempty"` // only on role=tool
 	ToolCalls  []ToolCall `json:"tool_calls,omitempty"`
 }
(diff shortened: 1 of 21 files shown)


jale-dbox:
## Final report

**What.** A generic, wire-neutral **structured-output** capability on the LLM provider layer: a caller can attach a JSON Schema to a `ChatRequest` and get back a guaranteed single JSON object using each provider's native mechanism. This is the platform half of a coordinated change (first consumer: the enterprise domain-pack generator, decisionbox-enterprise#232, which was failing on a self-hosted qwen quant behind LiteLLM→Ollama wrapping its output in a JSON array).

**How.**
- `ChatRequest.ResponseFormat` (`{Name, Schema, Strict}`) + `ProviderMeta.SupportsStructuredOutput` (surfaced in `/api/v1/providers/llm`; `GetProviderMeta` lets callers gate on it). Advisory — unsupported providers ignore it.
- Shared helpers `ApplyResponseFormatAsTool` / `NormalizeStructuredToolResponse` for the Anthropic wire (which has no `response_format`): force a single tool and fold its result back into `Content`, with `disable_parallel_tool_use` so exactly one object comes back.
- Per-provider: OpenAI-compat (`openai`, `litellm`, bedrock-openai) via `response_format` json_schema in the shared `BuildRequestBody`; Ollama via the `format` field (llama.cpp grammar); Anthropic (`claude`, bedrock-anthropic) via the forced tool. **Enabled** for ollama/litellm/openai/claude/bedrock; azure-foundry/vertex-ai stay off (their Claude/Gemini wires lack the tool plumbing) and clear the field so it's never forwarded upstream.

**Non-regression.** `ResponseFormat` is orthogonal to `Tools` (the synthetic tool is injected only when the caller supplied none), a nil `ResponseFormat` leaves every request byte-identical, and open-ended (dynamic-key) objects are preserved (non-strict json_schema; Ollama grammar + Anthropic tool schemas accept them). No new Go modules; `go.sum` unchanged.

**How verified.** `go build`/`go test`/`golangci-lint` green across every touched module; CI + CLA green. Tests: the forced-tool helper (inject/no-op/caller-tools-win/normalize/empty-object/multi-call-error/reserved-name), `response_format` wiring + omit-when-nil + open-map preservation, a hermetic Ollama `format` round-trip, an end-to-end Claude wire test (forced tool + `disable_parallel_tool_use` + fold-back), and per-provider capability-flag assertions. **6 Codex rounds** (empty-object fold, unsupported-provider forwarding, stop-reason normalization, reserved names, parallel-tool-use); the final round's suggestion was a theoretical edge on a caller path that doesn't exist, so that speculative handling was dropped to keep the change practical.

**Merge order.** Merge this first; enterprise#232 builds against it.

— Co-coded with Jale 🤖

abacigil (approved):

PR #343, opened 2026-08-27
fix(agent): harden discovery LLM-output parsing — recommendations (#342) + exploration action (#341)

### LINKED ISSUE #341: Harden discovery exploration action parser — fails with action="" on models that wrap the action in reasoning/prose
## Problem

Discovery **bricks** when the LLM doesn't emit a perfectly clean single JSON action object. Real failure (customer on the DecisionBox AI gateway):

```
discovery run failed: exploration failed: step 1: unable to parse LLM response after 4 attempts:
action JSON has no query, lookup_schema, search_tables, done flag, or recognized action (got action="")
```

The exploration loop (`services/agent/internal/ai/exploration.go`) requires the model to return **exactly one** JSON object with a recognized action key and *"no prose, no markdown fences, no planning JSON before the action"* (prompt at `exploration.go:647-650`). Flow: `ParseAction` (`:674`) → `extractJSON` (`:675`, defined `:837`) → `json.Unmarshal` (`:681`) → validate action keys (`:707-718`, error `:718`). When the model wraps the action in a `<think>` block, leading prose, markdown fences, or emits a "planning" JSON object first, extraction lands on the wrong object → `action=""` → all `maxParseRetries` exhaust (`:655`) → discovery dies at step 1.

Same class as pack-gen **#231** (structured-output adherence), but in the **discovery path**, which #231 didn't touch. It makes reasoning models / weaker models / some open models (e.g. Qwen) unusable for discovery. Current workaround: force a strict model (we switched the AI-gateway analysis alias to Opus 4.8).

## Root cause

- `extractJSON` (`:837`) doesn't strip `<think>…</think>` and doesn't robustly prefer the object carrying a **recognized action key** when multiple JSON objects are present — it can select a leading planning/thinking-only object → empty action.
- The corrective retry re-asks blindly; it doesn't tell the model *what* was wrong (empty/unrecognized action).
- The action isn't constrained via provider-native structured output/tool-use — **platform#340** shipped exactly that capability and it's unused here.

## Fix (reuse #231 / #340 helpers)

1. **Tolerant extraction** — strip `<think>…</think>` and markdown fences; scan **all** JSON objects in the response and pick the first carrying a recognized action key (`query_data` / `lookup_schema` / `search_tables` / `done` / `complete`), not merely the first/fenced one; unwrap single-element arrays. (`extractJSON:837`, `ParseAction:674`.)
2. **Targeted repair retry** — on parse/validation failure, feed the model its own output + the exact reason ("no recognized action — reply with ONLY `{\"query\": …}` or `{\"done\": …}`") instead of a blind re-ask. (retry loop `~:655`.)
3. **Schema-/tool-constrained action (preferred where supported)** — use platform#340's structured-output / forced tool-call to pin the action shape per provider (Ollama `format`, OpenAI `json_schema`/tool, Bedrock/Claude tool-use). Fall back to (1)+(2).
4. **Guard reasoning starvation** — ensure the exploration call's `max_tokens` leaves room after a `<think>` block so the action isn't truncated to empty.

## Acceptance criteria

- Responses containing a `<think>…</think>` block, leading prose, markdown fences, or a leading planning JSON **followed by** a valid action all parse successfully (unit tests over a fuzz corpus of these shapes).
- With a schema/tool-constrained provider, the action is always well-formed.
- A discovery run against a reasoning/open model (e.g. `qwen3-235b-a22b-instruct`) completes step 1 instead of failing with `action=""` after N retries.

## Evidence / refs

- Failure text above. Parser: `services/agent/internal/ai/exploration.go` — `ParseAction:674`, `extractJSON:837`, validation `:707-718`, retry `:655`, strict prompt `:647-650`.
- Related merged work to reuse: **enterprise#231** (pack-gen tolerant parse + repair + leaner examples), **platform#340** (schema-constrained structured output across providers).

### LINKED ISSUE #342: Recommendations silently zeroed: strict parse of expected_impact drops the whole batch when the LLM returns a string
## Problem

Discovery runs complete with insights but **0 recommendations — silently, across all LLM models**. The recommender generates good recommendations; the parse then discards the **entire batch** because of one field's type.

Real failure (redshift project, dev — captured in `discovery_recommendation_log`):

```
insight_count=1, response_len=3320, parsed_recommendations=0,
error: parse error: json: cannot unmarshal string into Go struct field
       Recommendation.recommendations.expected_impact of type models.Impact
```

The LLM returned a valid recommendation (title, description, implementation steps) but wrote **`expected_impact` as a prose string** (e.g. *"improves utilization accuracy and revenue forecasting"*) instead of the `Impact` object. `json.Unmarshal` of the whole `{"recommendations":[…]}` envelope (`services/agent/internal/discovery/orchestrator.go:1332`) then fails → the parse fn returns `[]` (`:1339`) → **0 recommendations persisted, no error surfaced to the user.**

## Why it's model-independent + high severity

- `Recommendation.ExpectedImpact` is a struct `Impact{metric, estimated_improvement, reasoning, …}` (`services/agent/internal/models/discovery.go:131`, type at `:145`) with **no custom `UnmarshalJSON`** → default strict decoding.
- **Any** model that describes impact in prose (natural + common) trips it — confirmed across multiple different LLMs, none of which was the previously-used model.
- **Silent:** the run reports success with N insights and 0 recommendations; the only trace is `recommendation_log.error`. Users see "no recommendations" with no reason.
- Same strict-JSON-parsing family as pack-gen **#231** and exploration **#341**, but worse: **one bad field on one recommendation drops the whole array.**

## Fix

1. **Tolerant `Impact.UnmarshalJSON`** (`models/discovery.go:145`) — accept **either** the `Impact` object **or** a bare string; coerce a string into `Impact{Reasoning: <string>}` (or a summary field). Mirror #231's tolerant DomainPack decoding.
2. **Don't fail the whole batch on one field** — decode recommendations **per-item** and skip/repair the offender instead of returning `[]` for the array (`orchestrator.go:1280-1339`); record drops in `RecommendationStep` telemetry (`recommendations_dropped*`).
3. **Prefer schema-constrained output** where the provider supports it (reuse **platform#340**) to pin the recommendation shape at generation time.
4. **Surface it** — a 0-recommendations result caused by a stashed parse error should log a WARN and set a step `status`/`error` the dashboard renders, instead of silently showing an empty section.

## Acceptance criteria

- A recommendation whose `expected_impact` is a plain string parses successfully (coerced), with a unit test.
- A batch containing one malformed recommendation keeps the valid ones instead of returning 0.
- Re-running the parser over the already-captured redshift dev response (`discovery_recommendation_log`, project `6a8ec022eb80c67bce27b911`) yields ≥1 recommendation.

## Evidence / refs

- `discovery_recommendation_log` for project `6a8ec022eb80c67bce27b911`: full LLM response present, `parsed=0`, unmarshal error above. Insight verdicts on that run: `{confirmed:1, partial:13, rejected:3, unverifiable:2}` → 1 eligible insight fed the recommender, which produced content that then failed to parse.
- Parse: `services/agent/internal/discovery/orchestrator.go:1332` (envelope struct), empty return `:1339`. Type: `services/agent/internal/models/discovery.go:131` (`ExpectedImpact Impact`), `:145` (`type Impact struct`). Confirmed **no** `Impact.UnmarshalJSON`.
- Related merged/open work to reuse: **#231** (pack-gen tolerant parse + repair), **#341** (exploration action parser hardening), **#340** (schema-constrained structured output).

jale-dbox:
## Summary

Hardens discovery's parsing of LLM output on two adjacent code paths that failed the same way — strict JSON decoding discarding good work when a model deviated from the exact expected shape. Combined into one PR (per request) because both fixes share new `ai.Client` structured-output plumbing and were verified together against a live project.

- **#342 — recommendations silently zeroed.** Runs completed with insights but **0 recommendations, across every model**, when the LLM emitted a field with the wrong JSON scalar type. Strict decoding of the `{"recommendations":[…]}` envelope failed the whole batch and returned `[]` with no user-visible reason.
- **#341 — discovery bricking at step 1.** The exploration action parser locked onto the wrong object when a model wrapped its answer in a `<think>` block, prose, or a planning object, decoded the action to empty, exhausted the retry budget, and killed the run.

Closes #341. Closes #342. (Supersedes the plan in #344, which is folded in here.)

## #342 — defense in depth

| Layer | Mechanism |
|-------|-----------|
| **Prevent** | Schema-constrained generation where the provider supports it (reuses #340): a curated response schema pins the recommendation shape at generation time. |
| **Tolerate** (load-bearing) | Tolerant decoding on every provider: `Impact.UnmarshalJSON` coerces a bare-string `expected_impact` into `Impact{reasoning}`; `Recommendation.UnmarshalJSON` coerces `priority` (accepts `"high"`/`"critical"`/`"P2"`), `segment_size`, and `confidence` from strings. Recommendations are decoded **per item** and a bare top-level array is accepted, so one malformed entry is skipped + counted, not fatal to the batch. |
| **Self-heal** | A bounded corrective re-prompt (`RECOMMENDATION_PARSE_MAX_RETRIES`, default 1) recovers a fully-unparseable response — fires only on a genuine parse failure, never on a partial success or a legitimately empty result. |
| **Observe** | `recommendation_parse_error` status + `recommendations_dropped_parse` / `recommendation_parse_retries` telemetry, mirrored to the `recommendation-log` API and rendered as the dashboard's empty-state reason. |

The two demonstrated failure modes (`expected_impact` as prose, `priority` as a word) and the wrapper cases (bare array) all came from **real captured `discovery_recommendation_log` responses** on the dev stack. The tolerant parse is load-bearing — schema-constraint is off on `vertex-ai`/`azure-foundry` and isn't a hard guarantee elsewhere, so the deterministic decode is what makes the fix hold everywhere.

## #341 — tolerant exploration action parsing

- **Strip reasoning blocks.** `extractJSON` removes `<think>`/`<thinking>` pairs (recovering an action written *inside* one via a full-text fallback), so JSON a model drafts in its reasoning can't be selected over the real action.
- **Recognise all action shapes.** `jsonHasActionKey` now recognises `lookup_schema`, `search_tables`, and the tool-use envelope (`{"name":…,"input":…}`) — so schema-discovery actions and envelopes win over trailing reasoning fragments (the direct cause of the customer `action=""` error).
- **Skip, don't abort, on an unbalanced brace.** A stray `{` in a preamble no longer stops the scan for a later valid action — while a nested action inside a *truncated* object (`{"plan":{"query":…}`) is correctly rejected so a draft query isn't run.
- **Reason-aware repair retry** and a per-step output budget sized to the model's **catalogued token cap** (was a hard-coded 4096), so a long reasoning block can't truncate the action to empty.
- **Structured output** on the action shape where the provider supports it.

## Shared plumbing

New `ai.Client` helpers — `SupportsStructuredOutput()`, `CreateMessageWithFormat`, `ChatWithFormat` — thread `ChatRequest.ResponseFormat` and self-gate on provider capability, so an unsupported provider transparently falls back to plain prompting. Existing `Chat`/`CreateMessage*` callers are unchanged (they funnel through one core that now carries an optional format).

## Verification

- **Real captured data (the strongest signal for #342):** replaying the four `discovery_recommendation_log` responses that had produced **0** recommendations now yields **3, 9, 9, 6** — covering the string `expected_impact`, string `priority`, and bare-array modes.
- **Live end-to-end run (#341)** against the dev project on the real warehouse + LLM gateway: at step 1 the model returned a prose preamble (`action=""`) — the exact brick — and the hardened parser caught it, sent the reason-aware nudge, recovered on the retry, and exploration ran to completion (9 steps). (The run didn't reach the recommendation phase within the time budget due to the slow shared gateway during the 30-insight validation phase — environmental, not a code path.)
- **Unit tests:** tolerant `Impact`/`Recommendation` decoding (string impact, priority words, numeric strings), per-item parse (array, skip-malformed, wrong-envelope-key, all-items-dropped telemetry), retry recovery/exhaustion, structured-output gating, drop-stat composition, schema drift guard; and for #341 a reasoning/action fuzz corpus (+ a no-action corpus), widened `jsonHasActionKey`, the skip-not-abort scan incl. the nested-draft rejection, `stripReasoningBlocks`, token-cap and structured-output gating.
- **Local gate:** `make build`, `make test-go`, `make lint-go` (0 issues), `make test-ui` (341 pass), `make lint-ui`, and `npm run build` all green. (One pre-existing test, `libs/go-common/mongodb.TestNewClient_UnreachableHost`, fails only in this shared dev environment because its hard-coded "unreachable" port is live here — unrelated to this change.)
- **Review:** Codex ran round after round until clean — 8 rounds of findings, each fixed in a new commit and re-reviewed, then a **clean 9th round with zero findings** (Rule 7). Highlights: wrong-key / null / capitalized envelopes treated as empty; unstamped parse-drop telemetry on total failure; a mislabelled drop reason in the live status; `"lowest"` priority mapping; `NaN`/`Inf` numeric strings that would poison the stored floats; a nested draft action extracted from a truncated `{"plan":…}`; an action inside `<think>` lost before recovery; a query containing a literal `<think>` tag being mutated by reasoning-stripping (→ moved to position-preserving masking that slices the original bytes); and the exploration output budget repeatedly risking context-window overflow (→ settled on a safe, configurable ceiling). Copilot pass to follow.

## Docs

`CHANGELOG` (Fixed), `docs/reference/data-models.md` (Impact coercion + new telemetry fields), `docs/reference/configuration.md` (`RECOMMENDATION_PARSE_MAX_RETRIES`), and `docs/concepts/discovery-lifecycle.md` (tolerant action parsing) updated.

— Co-coded with Jale 🤖


jale-dbox:
Per the review question about retry / self-healing — folded into the plan (**Fix 4**, §3/§4/§5/§7/§8/§9).

Two layers, framed on patterns already in the tree:
- **Deterministic self-heal (Fixes 1–2, no LLM cost):** coerce a string `expected_impact` → `Impact{reasoning}`, and per-item decode so one bad rec is skipped, not the whole batch. This is what actually fixes the reported bug, for every provider.
- **LLM re-prompt self-heal (Fix 4, last resort):** mirrors the exploration engine's existing `runStepWithRetry` (`ai/exploration.go:582`) — re-prompt on an unparseable response, bounded by `maxParseRetries`. Here it fires **only** when in-process parsing salvages zero recs from a non-empty response, so the happy/partial paths cost nothing extra. Config-driven cap `RECOMMENDATION_PARSE_MAX_RETRIES`, default **1**; `RecommendationParseRetries` telemetry.

Straight take: after the deterministic coercion the retry's marginal value is small — it earns its keep only on a fully-garbled envelope, and it does **not** fix genuine max-token truncation (that's the #140 token-budget path, which we log distinctly). Kept minimal (one batch-level retry, no per-item re-prompt, no `FixHistory` apparatus) to avoid over-building.

Still a plan — implementation follows approval.

— Co-coded with Jale 🤖

jale-dbox:
Reworked the plan to do this **properly** — defense-in-depth, four layers, which is the issue's own fix list in full + the retry:

| # | Layer | Mechanism | Coverage |
|---|-------|-----------|----------|
| 1 | **Prevent** | schema-constrained generation (#340) | claude/bedrock/openai/ollama/litellm |
| 2 | **Tolerate** *(load-bearing)* | tolerant `Impact.UnmarshalJSON` + per-item decode | **every** provider |
| 3 | **Self-heal** | bounded re-prompt (`runStepWithRetry` pattern) | mainly vertex-ai/azure-foundry |
| 4 | **Observe** | drop/retry telemetry + dashboard reason | all |

Two calls I'd stand behind:
- **Layer 2 stays load-bearing.** Schema-constraint is prevention, not a guarantee — it's off on vertex-ai/azure-foundry and a "supported" model can still drift. The deterministic parse is what makes the fix hold everywhere, so it's the base commit; Layer 1 is additive on top.
- **Curated schema, not reflected.** Hand-author the input-contract subset (no `id`/`created_at`/`validation`/`description_md` leaking into the generation contract), guarded by a reflection-based drift test (schema keys ⊆ struct tags) — a struct rename then fails CI. Avoids a new dependency and gives the model a cleaner contract.

Full detail in the updated description (§3 Fix 5, §4, §7, §8, §9). Still a plan — implementation on approval.

— Co-coded with Jale 🤖

jale-dbox:
## Final report

**What.** One PR hardening discovery's parsing of LLM output on two adjacent paths that failed the same way — strict JSON decoding throwing away good work when a model deviated from the exact shape. Closes **#342** (recommendations silently zeroed) and **#341** (discovery bricking at step 1). The #341 plan PR (#344) is folded in and closed.

**How.** Defense in depth.
- *#342:* tolerant decoding on every provider (`Impact.UnmarshalJSON` coerces a prose-string `expected_impact`; `Recommendation.UnmarshalJSON` coerces `priority`/`segment_size`/`confidence` from strings), per-item decode that also accepts a bare top-level array, schema-constrained generation where supported (reuses #340), a bounded corrective re-prompt, and `recommendation_parse_error` status + `recommendations_dropped_parse`/`recommendation_parse_retries` telemetry surfaced to the API and dashboard.
- *#341:* mask (not strip) `<think>` reasoning so drafts inside it are ignored while a query containing a literal `<think>` tag is never mutated; recognise `lookup_schema`/`search_tables`/tool-envelope actions; reject truncated/malformed objects (→ retry) instead of extracting a nested draft; a reason-aware repair nudge; a configurable per-step output budget; and structured output on the action shape.
- Shared: new `ai.Client` `SupportsStructuredOutput` / `CreateMessageWithFormat` / `ChatWithFormat`, self-gating on provider capability.

**Why.** Runs were completing with insights but **0 recommendations across every model** (a single wrong-typed field failed the whole batch), and reasoning/open models **bricked exploration at step 1** when they wrapped the action in prose or `<think>`.

**How it was verified.**
- **Real captured data** — replaying the four `discovery_recommendation_log` responses that had produced **0** recommendations now yields **3, 9, 9, 6** (covering string `expected_impact`, string `priority`, and bare-array modes). This is the strongest signal and it drove a fix the synthetic tests missed (string `priority`).
- **Live discovery** against the dev project on the real warehouse + LLM gateway: at step 1 the model returned a prose preamble (`action=""`) — the exact brick — and the hardened parser caught it, sent the reason-aware nudge, recovered on the retry, and exploration ran to completion (9 steps). (It didn't reach the recommendation phase within the time budget due to the slow shared gateway in the 30-insight validation phase — environmental, not a code path.)
- **Unit/integration tests** across models, discovery, ai, and the API handler (per-item parse incl. wrong-key/null/capitalized envelopes, retry recovery/exhaustion, structured-output gating, drop-stat composition, schema drift guard, NaN/Inf rejection; and the #341 reasoning/action fuzz corpus + no-action corpus, mask preservation, truncated-nested rejection, token budget).
- **Local gate + CI:** `make build`, `make test-go`, `make lint-go` (0 issues), `make test-ui` (341 pass), `make lint-ui`, `npm run build` — all green; **CI green on the merge commit** (Go/Dashboard build+test+lint, License, Docs Lint, CodeQL).

**Review rounds.** Codex ran **9 rounds** — 8 with findings (each fixed in a new commit and re-reviewed), then a **clean 9th round with zero findings** (Rule 7). It caught, among others: wrong-key/null/capitalized envelopes read as empty; unstamped parse-drop telemetry on total failure; a mislabelled live-status drop reason; `"lowest"` priority mapping; `NaN`/`Inf` strings that would poison the stored floats; a nested draft extracted from a truncated `{"plan":…}`; an action inside `<think>` lost before recovery; a query with a literal `<think>` tag being mutated by reasoning-stripping; and repeated context-window-overflow risk in the exploration token budget. Copilot did **not** post a review (it can't be requested via API here and hasn't auto-reviewed the recent PRs on this repo).

**Follow-ups / known limitations.**
- Adding this PR to project board 4 failed — the bot token lacks project write scope (`Resource not accessible by personal access token`); needs a manual add or a token with the project scope.
- `EXPLORATION_MAX_OUTPUT_TOKENS` defaults to 4096 (the proven-safe value); deployments running verbose reasoning models on large-context backends should raise it. A fuller fix would plumb the project's effective context window (e.g. Ollama `num_ctx`) into the exploration budget — out of scope here.
- Two repo tests fail **only in this shared dev container** because they assume env/ports it overrides (`mongodb.TestNewClient_UnreachableHost` hard-codes a port that's live here; `config.TestLoad_MissingMongoDBURI` assumes `MONGODB_URI` is unset) — both pass in CI, which is green.

— Co-coded with Jale 🤖

abacigil (approved):

Issue #356, opened 2026-08-31
Telemetry for multi-datasource routing: routing decisions, refused hops, dropped filters, source quality flags


jale-dbox:
> **PR base branch: `feat/non-sql-datasources`** (not `main`).
> This is the shared bridge branch for a multi-issue, cross-repo feature. Bring your
> branch up to date with it (`git fetch origin feat/non-sql-datasources && git rebase
> origin/feat/non-sql-datasources`) and open the PR with
> `gh pr create --base feat/non-sql-datasources`. CI already runs on that branch.

## Goal

Multi-datasource routing fails in a way that never arrives as a bug report: the output is **well-formed, confident, and wrong**. A question routed to the wrong datasource, a cross-source hop that quietly widened its filter, or a result the source itself flagged as partial all produce a clean-looking answer. None of it is currently visible.

Emit the telemetry that makes those failures observable.

## Scope

Instrument, at minimum:

| Event | Carries |
|---|---|
| **Routing decision** | question, candidate datasources, chosen one, confidence, whether the turn clarified instead |
| **Refused hop** | a cross-source hop rejected rather than executed, and why |
| **Dropped filter** | a requested filter that could not be expressed in the target's query language |
| **Source quality flag** | a result the source reported as sampled, thresholded, or truncated |

- Follow the existing telemetry/event conventions in `services/agent` — no new transport, no new sink.
- The events must be attributable to a run and a datasource, or they are not actionable.
- Emission only.

## Acceptance

- [ ] Each event above is emitted at the point the decision is made, with the datasource and run identifiable.
- [ ] A routed Ask turn and a discovery run both produce the routing events end to end (covered by a test, not by eyeballing logs).
- [ ] No PII and no raw customer data in event payloads — question text handled per existing convention.
- [ ] `make test` / `make lint` green.

## Out of scope

- Dashboards, alerting, or aggregation on the telemetry — emission only.
- Anchoring-refusal events: the anchoring concept does not exist in the platform yet; those land with the issue that introduces it.

## Notes

Deliberately independent — this can land before the seam work and does not block on it.

Issue #4, opened 2026-03-15
fix(helm): UID mismatch — Dockerfile uses 1000, Helm uses 10001


abacigil:
## Context
PR #2 review finding **C2**.

Helm sets `runAsUser: 10001` / `runAsGroup: 10001`, but Dockerfiles create users with UID 1000. All files in the images are owned by UID 1000. When K8s enforces UID 10001, the process can't read its own binary, domain packs, or Next.js output → **permission denied on everything**.

## Locations
- `helm-charts/decisionbox-api/values.yaml`
- `helm-charts/decisionbox-dashboard/values.yaml`
- `services/api/Dockerfile`
- `ui/dashboard/Dockerfile`

## Fix
Either change Helm to `runAsUser: 1000` / `runAsGroup: 1000`, or change Dockerfiles to use UID 10001. Recommend aligning on UID 1000 (standard non-root).

## Reference
- PR2-REVIEW.md — C2

PR #361, opened 2026-08-31
feat(askserve): record the router's ballot and verdict, not just what the turn queried

jale-dbox:
Closes #356. Part of #163 (epic). Targets the bridge branch `feat/non-sql-datasources`.

## What the issue asked for, and what the code actually supports

The issue listed four event kinds. Tracing each against the code first:

| Event | State | Where it belongs |
|---|---|---|
| **Routing decisions** | Partly present — reason, confidence, clarify and the queried datasources are already recorded and persisted | **This PR completes it** |
| Refused hops | **No call site exists** — nothing refuses a hop today | The join-key / anchoring work, which introduces refusal |
| Dropped filters | **No call site exists** — nothing drops a filter today | The adapter + join-key work |
| Source quality flags | **No call site exists** — no source reports sampling/thresholding today | The GA4 adapter, which is the first source that can report one |

Emitting event types for three concepts that do not exist would mean writing instrumentation with no producer, which the issues that introduce those concepts would then have to rework to fit what they actually learn. Those three are left to land with the code that can raise them. Recorded here so the omission is a decision rather than an oversight.

## The gap that is real, and fixable now

Routing telemetry records the router's reason, confidence, clarify flag, and the datasources the turn went on to **query**. It does not record what the router chose **from**.

That distinction is load-bearing for this epic. Today, a turn that ignored a connected datasource and a turn where that datasource was correctly irrelevant produce **identical records** — both just show the other datasource being queried. So "the router never picks this source, for any question" is invisible in the data, and stays invisible however many turns you look at.

That is precisely the failure the epic flags as must-fix #4: a source whose retrieval evidence is weak contributes nothing to routing, for every question, and nothing errors. It is also what a routing evaluation set has to measure against.

## The change

Two fields — the datasources the router chose **from** (the ballot) and the ones it **chose** (the verdict) — carried through `TurnFinal` to the persisted turn record, alongside the existing fields.

Four details, each pinned by a test:

- The **ballot is recorded before the model is asked**, so a router error leaves a record of what it was looking at rather than nothing at all.
- On a **clarify**, the ballot is recorded with an empty verdict. This is the case it matters most: nothing is chosen and nothing is queried, so without it the record shows only that a question was asked.
- The **verdict is the post-validation set**, so an id the model invents never reaches the record.
- A turn with **no routing decision records neither** — the fields mean "the router ran and saw this", and filling them on a pinned single-datasource turn would invent a decision that never happened.

## Testing

`go test ./...` green for `services/agent` and `libs/go-common`; `golangci-lint run` 0 issues on both; `codex review` no findings.

Verified non-vacuous by mutation: dropping the ballot assignment fails the new tests.

### DIFF
--- a/libs/go-common/models/ask_turn.go
+++ b/libs/go-common/models/ask_turn.go
@@ -69,6 +69,16 @@ type AskTurn struct {
 	// empty/zero for a single-datasource project or a turn pinned to an
 	// explicit datasource, where no routing decision is made.
 	RoutedDatasourceIDs []string `bson:"routed_datasource_ids,omitempty" json:"routed_datasource_ids,omitempty"`
+	// RoutingCandidateIDs are the datasources the router chose FROM, and
+	// RoutingChosenIDs the ones it chose. Both are distinct from
+	// RoutedDatasourceIDs, which records what the turn went on to query.
+	//
+	// The distinction is the whole point: without the candidate set, a turn
+	// that ignored a connected datasource is indistinguishable from one where
+	// that datasource was correctly irrelevant. Answering "is the router blind
+	// to this source?" needs to know it was on the ballot and lost.
+	RoutingCandidateIDs []string `bson:"routing_candidate_ids,omitempty" json:"routing_candidate_ids,omitempty"`
+	RoutingChosenIDs    []string `bson:"routing_chosen_ids,omitempty" json:"routing_chosen_ids,omitempty"`
 	RoutingReason       string   `bson:"routing_reason,omitempty" json:"routing_reason,omitempty"`
 	RoutingConfidence   float64  `bson:"routing_confidence,omitempty" json:"routing_confidence,omitempty"`
 	RoutingClarify      bool     `bson:"routing_clarify,omitempty" json:"routing_clarify,omitempty"`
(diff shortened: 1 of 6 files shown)


jale-dbox:
Ran a Codex review pass over this PR's merge diff against `feat/non-sql-datasources` (it hadn't been reviewed since it was opened on Aug 31).

**Result: clean.** Verbatim:

> No actionable defects were found in the merge diff. The new routing candidate/chosen telemetry is carried through turn state and persistence consistently.

Current state, for whoever picks this up:

| | |
|---|---|
| CI | all green — Go Build / Test / Lint, License Check, codecov/patch, cla |
| Mergeable | `MERGEABLE` / `CLEAN` |
| Unresolved review threads | 0 |
| Behind the base | 0 commits |
| Human reviews | **0** — this is the only thing missing |

**Merge this before #362.** #362 (`feat/anchoring-resolution`) is stacked on this branch and carries its two commits underneath its own, so merging #362 first would close this one as already-merged rather than reviewed on its own.

Both are blockers for the enterprise-side work that reads the anchoring flag and emits the refusal telemetry.

— Co-coded with Jale 🤖

PR #362, opened 2026-08-31
feat(warehouse): resolve a datasource's effective anchoring, demotion only

jale-dbox:
Part of #163 (epic), the platform half of the anchoring work (decisionbox-io/decisionbox-enterprise#242, W6). Targets the bridge branch `feat/non-sql-datasources`.

This is **resolution only**. Enforcement — refusing a non-anchoring source as a project's only datasource, and refusing standalone Discovery — lands next in the enterprise handler, on top of this.

## The rule

Anchoring answers two different questions that compose one way only:

- **`CanAnchor`** (provider) — what a source of this *type* is capable of. Shipped with the capability descriptor.
- **`Anchoring`** (datasource) — what *this customer's* source is being used as. Added here.

`EffectiveAnchoring(provider, override)` applies the provider as a **ceiling**:

| Provider | Override | Effective |
|---|---|---|
| anchors | unset | ✅ |
| anchors | `false` | ❌ (demoted) |
| anchors | `true` | ✅ |
| **cannot anchor** | unset | ❌ |
| **cannot anchor** | `true` | ❌ — **promotion never applies** |
| cannot anchor | `false` | ❌ |

Promotion is refused because it asserts something about the data that isn't true, and the standalone Discovery run it would license produces exactly the shallow output this feature exists to avoid. Demotion is always legal — "our warehouse already holds this, treat it as enrichment" is a true statement about where the records live.

`AnchoringOverrideAllowed` exists so a caller can **refuse** a promotion rather than store a value that silently does nothing, leaving the user believing they changed something.

## An unregistered provider resolves to anchoring

Deliberate, and the safer of the two options. A binary that has not linked a provider cannot construct it either, so such a datasource cannot be connected at all — the project is broken visibly, long before anchoring is consulted.

The opposite default would make anchoring depend on which providers a binary happens to blank-import: a build missing an import would quietly treat a legitimate warehouse as enrichment-only, with nothing at the call site to suggest why.

## The field

`WarehouseConfig.Anchoring *bool` on both the API and agent models. Nil means "not set" — the provider decides.

Deliberately **not** called `primary`: a project's primary datasource is a selection among those connected, which is a different concept already in use in this struct.

Callers resolve via `warehouse.EffectiveAnchoring(Provider, Anchoring)` rather than reading the field, so the provider ceiling is always applied. The doc comment on the field says so.

## Testing

Tests register real providers in the registry (both anchoring and not) rather than stubbing the lookup, so the ceiling is exercised through the same path production uses. Every row of the table above is a case.

`go test ./...` green for `libs/go-common`, `services/api`, `services/agent`; `golangci-lint run` 0 issues on all three.

### DIFF
--- a/libs/go-common/warehouse/anchoring_test.go
+++ b/libs/go-common/warehouse/anchoring_test.go
@@ -1,11 +1,21 @@
 package warehouse
 
-import "testing"
+import (
+	"fmt"
+	"sync/atomic"
+	"testing"
+)
+
+// probeSeq keeps each probe registration unique. The registry is package-global
+// and RegisterWithMeta panics on a duplicate name, so fixed slugs would panic
+// on the second iteration under `go test -count=2`.
+var probeSeq atomic.Int32
 
 // anchoringProbe registers a provider with a given CanAnchor value under a
 // unique slug, so these tests exercise the real registry rather than a stub.
-func anchoringProbe(t *testing.T, slug string, canAnchor bool) string {
+func anchoringProbe(t *testing.T, name string, canAnchor bool) string {
 	t.Helper()
+	slug := fmt.Sprintf("%s_%d", name, probeSeq.Add(1))
 	RegisterWithMeta(slug, func(ProviderConfig) (Provider, error) {
 		return &mockWarehouseProvider{}, nil
 	}, ProviderMeta{
--- a/libs/go-common/warehouse/anchoring_test.go
+++ b/libs/go-common/warehouse/anchoring_test.go
@@ -1,0 +1,75 @@
+package warehouse
+
+import "testing"
+
+// anchoringProbe registers a provider with a given CanAnchor value under a
+// unique slug, so these tests exercise the real registry rather than a stub.
+func anchoringProbe(t *testing.T, slug string, canAnchor bool) string {
+	t.Helper()
+	RegisterWithMeta(slug, func(ProviderConfig) (Provider, error) {
+		return &mockWarehouseProvider{}, nil
+	}, ProviderMeta{
+		Name:       slug,
+		Dialect:    "Probe SQL",
+		Capability: Capability{CanAnchor: Anchoring(canAnchor)},
+	})
+	return slug
+}
+
+func TestEffectiveAnchoring(t *testing.T) {
+	anchor := anchoringProbe(t, "probe_anchor_yes", true)
+	enrich := anchoringProbe(t, "probe_anchor_no", false)
+
+	tests := []struct {
+		name     string
+		provider string
+		override *bool
+		want     bool
+	}{
+		{"anchoring provider, no override", anchor, nil, true},
+		{"anchoring provider demoted", anchor, Anchoring(false), false},
+		{"anchoring provider explicitly kept", anchor, Anchoring(true), true},
+
+		// The ceiling: a provider that cannot anchor stays non-anchoring
+		// whatever the datasource says. Promotion would assert something about
+		// the data that is not true.
+		{"non-anchoring provider, no override", enrich, nil, false},
+		{"non-anchoring provider cannot be promoted", enrich, Anchoring(true), false},
+		{"non-anchoring provider demoted stays demoted", enrich, Anchoring(false), false},
+
+		// Unregistered providers resolve to anchoring — see the doc comment.
+		{"unregistered provider", "probe_never_registered", nil, true},
+		{"unregistered provider demoted", "probe_never_registered", Anchoring(false), false},
+	}
+	for _, tc := range tests {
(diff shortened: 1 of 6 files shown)


jale-dbox:
Ran a Codex review pass over this PR's merge diff against `feat/non-sql-datasources` (it hadn't been reviewed since Aug 31). **One finding, P2.** I have not pushed a fix — it's a scope call that belongs to a human, and I explain why below.

### [P2] A project can be saved with *no* anchoring datasource

`AnchoringOverrideAllowed` is the only anchoring validation reached from `Create` and from the legacy single-warehouse `Update`, and it rejects only the illegal direction — promotion of a `CanAnchor:false` provider. A request whose **sole** datasource sets `anchoring:false` passes: `AnchoringOverrideAllowed(…, false)` returns true, and the create path won't let a second datasource be added. The project persists with every datasource resolving to `EffectiveAnchoring == false`.

That contradicts the contract this PR establishes. Demotion is meant to say "this customer's records live elsewhere" — it is a statement about *which* source anchors, not permission for a project to anchor on nothing. A project in that state can still be asked to run a discovery, and there is no source in it that can carry one.

`services/api/internal/handler/projects.go:737`.

### Why I stopped at reporting

The fix is a project-level "at least one effective datasource anchors" check before save, which is straightforward — but it is a **behaviour change at the API boundary for existing projects**, and it overlaps the *"a non-anchoring source cannot be a project's only datasource"* acceptance criterion in decisionbox-io/decisionbox-enterprise#252, which may deliberately be a later PR in the sequence. Both readings are defensible:

- **Accept as scoped-out** — this PR resolves *effective anchoring* and refuses promotion; the only-datasource refusal is the enforcement half of W6 and lands with the discovery-launch guard.
- **Add it here** — the gap is reachable through this PR's own new field, so the PR arguably owns it.

I'd take the first, provided the enforcement half is genuinely next rather than assumed. Your call.

### Otherwise ready

| | |
|---|---|
| CI | all green — Go Build / Test / Lint, License Check, codecov/patch, cla |
| Mergeable | `MERGEABLE` / `CLEAN` |
| Unresolved review threads | 0 |
| Behind the base | 0 commits |
| Human reviews | **0** |

**Merge #361 first** — this branch is stacked on it and carries its two commits (`df80467`, `949456d`) underneath its own three. #361 came back clean from the same review pass.

— Co-coded with Jale 🤖

jale-dbox:
Rebased onto the bridge now that #361 is merged.

#361 was **squash-merged** as `d2a8b53`, so its original two commits (`df80467`, `949456d`) were not ancestors of the bridge — the merge-base still predated them, and this PR's diff was showing #361's changes a second time. Anyone reviewing it would have read the routing-telemetry work again.

**Before:** `+621/−29` across 12 files, 5 commits.
**Now:** `+302/−0` across 6 files, 3 commits — only the anchoring change.

```
1ecf076 test(api): give the anchoring probe providers pricing metadata
bb689ce fix(api): refuse an anchoring promotion instead of storing one
b5f125b feat(warehouse): resolve a datasource's effective anchoring, demotion only
```

`df80467` was skipped as already upstream and `949456d` was dropped by git itself ("patch contents already upstream"). No content of this PR's own was touched. Verified before pushing: `libs/go-common` and `services/api` build, `libs/go-common/warehouse` and the handler's anchoring tests pass, `services/agent` builds.

The [P2 finding above](https://github.com/decisionbox-io/decisionbox-platform/pull/362#issuecomment-5495113882) still stands and is unaffected by the rebase — a project whose sole datasource is demoted still saves with nothing that can anchor it. Your call whether that lands here or with W6's enforcement half.

— Co-coded with Jale 🤖

jale-dbox:
**Decision on the P2: accepted as scoped-out.** Confirmed by the team, so this PR is not blocked on it.

To restate what that means for the record, since the finding stays visible above:

This PR resolves a datasource's **effective anchoring** and refuses the illegal direction — promoting a provider that declares `can_anchor: false`. That is a complete, coherent piece of work.

What it deliberately does **not** do is the project-level check that at least one datasource still anchors after a demotion. So a project whose sole datasource sets `anchoring:false` will save, and end up with nothing that can carry it. That is the *enforcement* half of W6 in decisionbox-io/decisionbox-enterprise#252 — the same place the "a non-anchoring source cannot be a project's only datasource" and "cannot run standalone Discovery" refusals live — and it lands with them rather than here.

The one thing worth holding onto: the enforcement half now has a concrete gap waiting for it, not just an acceptance-criteria line. Whoever picks up W6 should start at `services/api/internal/handler/projects.go:737`.

No code change from the review. Rebased earlier onto the post-#361 bridge; CI green.

— Co-coded with Jale 🤖

PR #363, opened 2026-09-01
feat(warehouse,agent): carry source-reported result-quality caveats

jale-dbox:
## Why

Some sources answer a query with data that is **not a faithful answer to it**, and say so only in metadata alongside the rows:

- rows **withheld** below a reporting threshold,
- values **sampled** and extrapolated,
- a high-cardinality tail **truncated** into a catch-all bucket,
- fields **restricted** from the credential in use.

In every one of those cases the query *succeeds* and the rows come back well-formed. Nothing about the result distinguishes a sound one from an unsound one by inspection — so a share, total or ranking computed over withheld rows is a **wrong** answer, not a partial one, and downstream it becomes a confident number nobody has reason to doubt.

Today a provider that knows this has nowhere to put it. `QueryResult` is `{Columns, Rows}`, so the caveat is dropped at the provider boundary.

## What changed

- **`warehouse.QueryCaveat` channel** — `QueryResult.Quality []QualityCaveat`, with four degradation kinds (`withheld` / `sampled` / `truncated` / `restricted`) and a human-readable `Detail`, plus `(*QueryResult).Degraded()`.
- **`ExecuteResult.Quality`** — the executor copies the caveats off the result that actually succeeded, so they survive the repair loop to the caller.
- **A warn-level log** naming what was degraded, for the case where the caller does nothing with the field yet.

Caveats are **advisory, not errors**. A degraded result is still an answer; treating a caveat as a failure would throw away a usable one. The caller decides what it means for the conclusion it is about to draw — which it can only do if the caveat reaches it.

## Nothing changes for SQL

Every SQL warehouse answers exactly what was asked or fails, so none declares a caveat, the field stays nil on every existing path, and the executor invents none (asserted — a non-nil empty slice would read as "the source said something"). The **characterization tests covering the repair loop are green and unchanged**.

## Verification

- `libs/go-common/warehouse` — `Degraded()` across nil receiver / zero value / rows-without-caveat / explicitly-empty slice / caveat-on-empty-rows; `String()` rendering incl. the detail-less case; a guard that the four kinds stay distinct and non-empty (they are persisted and branched on, so a collapse would silently merge two classes).
- `services/agent/internal/queryexec` — caveats reach the caller with the rows intact; a sound result leaves `Quality` nil; **on the repair path the caveat is the succeeding attempt's**, not a stale one; a fully-failed query reports none.
- `go build` + full `queryexec` and `warehouse` suites green.

## Note for the reviewer

This is the seam side of a silent-failure class. It is deliberately narrow: it adds a channel and proves the SQL path is untouched. Nothing in this PR *populates* it — the first producer is an adapter in the plugin repo, and carrying the caveat further (into stored steps and the insight envelope) is separate work that follows the routing/telemetry PRs.

— Co-coded with Jale 🤖


jale-dbox:
Context on why this one is worth merging promptly, since it looks like a standalone type addition.

**It is the blocking dependency for [decisionbox-enterprise#253](https://github.com/decisionbox-io/decisionbox-enterprise/pull/253)** (the first non-SQL datasource adapter). That PR's CI is currently red on exactly this and nothing else:

```
datasources/.../quality.go:27: undefined: gowarehouse.QualityCaveat
datasources/.../provider.go:351: unknown field Quality in struct literal of type warehouse.QueryResult
```

Enterprise CI resolves the platform branch from `.platform-ref` (`feat/non-sql-datasources`), so it goes green the moment this lands on the bridge.

The adapter is the first *producer* of these caveats — it reads them from a source that returns HTTP 200 with complete-looking rows while having withheld, sampled or truncated the data behind them. Without this channel the caveat is discarded at the provider boundary and a number computed from withheld rows becomes a confident wrong answer downstream, which is the failure this type exists to make impossible.

Nothing in *this* PR populates the field, and the SQL path is asserted unchanged (characterization tests green and unchanged), so it should be a short read.

— Co-coded with Jale 🤖

PR #366, opened 2026-09-01
feat(agent): describe each datasource in the language it actually accepts

jale-dbox:
## Why

The Ask turn was SQL-shaped everywhere, and said so in a dozen places across both model paths:

| where | what it said |
|---|---|
| system prompt, opening line | "running read-only **SQL** against their data warehouse" |
| JSON action vocabulary | `{"query":"SELECT ...","purpose":"..."}` |
| TOOLS / RESULT HANDLING | "run one read-only **SQL** query", "write aggregate **SQL** (COUNT, SUM, AVG, GROUP BY)" |
| GROUNDING | "your FIRST action must be a discovery query — e.g. `SELECT table_name FROM <dataset>.INFORMATION_SCHEMA.TABLES`" |
| `query_data` description | "Run one read-only **SQL** query (SELECT / CTE only) against the data warehouse" |
| `search_tables` description | "search the indexed schema for **tables**", "Max number of **tables** to return" |
| the grounding nudge | "discover them with a query against **INFORMATION_SCHEMA** … or use lookup_schema" |
| tenant scope | "the predicate `x = 'y'` (in the **WHERE clause**, or preserved through every **join/CTE**)" |

Handed all of that about a source that has **no tables**, a model does not stop and ask. It resolves the contradiction in favour of the most concrete instruction — the tool description, which arrives attached to the very argument it governs — writes SQL, and spends the turn on a query the source was always going to reject. The failure then surfaces as a confusing query-time error rather than an honest "this source can't do that".

The capability descriptor already carried the answer (`Capability.Shape` / `QueryLanguage`, landed earlier on this branch). Nothing was reading it here.

## What changed

Every one of those statements is now **branched** on the datasource's declared shape, on both the native tool-calling path and the JSON-text fallback.

**A cube-shaped source** gets its own block, which states the absence **before** the language on purpose: a model that has read *"NO TABLES. There is nothing to SELECT from"* cannot then write a `FROM` clause by reflex, whereas one that reads a language name first fills the gaps with SQL habits.

```
DATASOURCE
- This source has NO TABLES. There is nothing to SELECT from and no schema to look up.
- It is a metric/dimension cube: you choose what to measure and what to break it down by,
  and the source computes the result.
- Queries are written as: Report Request (JSON) — not SQL, which this source rejects.
- Use search_tables to find the metrics and dimensions this source offers — each result says
  whether it is a metric or a dimension. Do not call lookup_schema against this source; it has
  no columns.
- The source is READ-ONLY.
```

**`search_tables` stops calling itself a table search.** It is the only discovery tool that works against such a source, and the prompt sends a cube turn straight to it — so a model told by the tool that it searches tables, and by the prompt that its source has none, had been given a reason not to call the one tool that would have worked. On a native provider the description drives tool *selection*, so this was not cosmetic.

**`lookup_schema` is withheld entirely** when nothing reachable has columns — from the tool set and from the JSON-text action list alike (the text parser accepts any action the list advertises, and `execLookup` runs it). This one matters because *while the turn is ungrounded the model is forced to call some tool*, so an advertised-but-impossible tool can consume the very step meant to gather evidence. It stays offered on a **mixed** turn, where it is still right for the SQL datasource — hence two facts rather than one (`sourceShapes{anyCube, allCube}`), pinned apart by a test.

**The grounding nudge** — sent when the model tries to answer before running anything — is shape-aware too, and it is the sharpest case: it arrives *after* the model has gone wrong, as the most recent thing in the conversation, and on the tool path as an error attached to the very call it is correcting. Sending a cube turn back to `INFORMATION_SCHEMA` from there outranks everything the system prompt said, at the moment it is most likely to comply.

**Tenant scope** on a cube source is stated as the source's own filtering rather than as a `WHERE` predicate over a column it does not have. The rule is unchanged and still stated in full.

**The branch reads what the turn can reach, not what it lists.** The router narrows the visible set, but `resolveQueryDatasource` validates a model-chosen `datasource_id` against the whole project on purpose (the router is a soft prior) and `search_tables` spans every datasource — so a cube the router passed over is one search result away. It is therefore also named at the end of the catalog, marked as *not selected for this question*: that keeps the router's prior instead of promoting the source into its choice, while making the catalog complete about what can actually be queried. A **SQL** datasource the router passed over stays hidden — the blanket guidance already describes it — and both halves are pinned by tests.

**Deliberately not changed:** nothing here specifies a non-SQL source's *request format*, because nothing in the turn supplies one — `search_tables` returns a source's metrics and dimensions, not its request envelope. Describing a format the prompt cannot show would be worse than naming the language and letting the model use what it knows about it. Tracked as a follow-up.

**Note on enforcement, unchanged here:** `verifyFilter` is a substring check on query text for SQL and non-SQL alike — it catches a request that never mentions the scope field and cannot verify what one that does actually asks for. Real scoping rests on the scoped credential and the governance middleware, as before; a structured-*payload* query is refused outright rather than checked.

## Nothing is generalised — only branched

A project whose datasources are all SQL — **every project that exists today** — renders byte-for-byte as it did before. That is a measurement, not an assertion: eight surfaces (both prompt paths × single/multi, the warehouse block, both tool definitions) were rendered under the base-branch implementation and under this one and compared byte-for-byte, re-run after every review round. The grounding nudge is pinned the same way, against its historic literal.

```
surfaces: 8 | DIFFERING: NONE
```

## Verification

- `make build`, `make lint-go` (all modules), `make lint-docs` — clean.
- `make test-go`: all modules pass except one **pre-existing environment artefact** unrelated to this change — `libs/go-common/mongodb TestNewClient_UnreachableHost` expects `localhost:27099` to be unreachable, and something in this container listens on it. That package is untouched by this branch and the test fails identically on the base commit.
- New tests (`prompt_cube_test.go`, plus one in `runtime_test.go`), aimed at the failure cases:
  - **`TestSQLOnlyTurn_SaysExactlyWhatItAlwaysSaid`** / **`TestCubeTurn_DropsEverySQLAssertion`** — every historic SQL sentence survives on a SQL turn and none survives on a cube turn. A single leftover is enough to lose the turn.
  - **`TestTurnRouting_Shapes_ReadWhatTheTurnCanReach`** / **`..._DistinguishAnyFromAll`** — the reachable set (including the routed-but-targetable case) and the any-vs-all distinction, in both directions.
  - **`TestToolsForPhase_WithholdsLookupSchemaWhenNothingHasTables`**, **`TestToolsPrompt_OffersOnlyToolsTheTurnWasGiven`**, **`TestTextPrompt_AdvertisesNoLookupWhereNothingHasTables`** — the prompt and the tool set agree on both paths.
  - **`TestGroundingNudge_CorrectsInTheShapeOfTheSource`**, **`TestTenantScope_IsStatedInTermsTheSourceCanHonour`**, **`TestDatasourcesSection_NamesAReachableCubeTheRouterPassedOver`**, **`TestLoop_OffersCubeShapedToolsToTheModel`** (asserts what the provider actually received, closing the outermost wiring).
- **Every branch was mutation-checked** — reverted in turn to its pre-change text, suite re-run. **46 distinct mutants; 45 killed.** Three survived on first attempt: two are the reason two of the tests above exist (the loop's own tool wiring, and `hiddenCubes` surfacing SQL datasources) and were killed after; the third — stripping the project list off the routing the prompt is built from — genuinely changes nothing observable, because `reachable()` falls back to the visible set and the two coincide on a single-datasource turn. That wiring is covered instead by `TestResolveTurnRouting_CarriesTheWholeProject`.
- **Codex: 8 rounds** — 7 findings, all fixed, then a clean round. It caught, in order: the reachable-vs-visible set, `search_tables`'s description, `lookup_schema` on the tool path, then on the text path, then the prompt still naming it, the grounding nudge, the tenant-scope wording, and the hidden reachable cube. Every one was a real contradiction between what the model was told and what it was given.

— Co-coded with Jale 🤖


jale-dbox:
## Final report

**What.** The Ask turn now describes each datasource in the language it actually accepts, instead of asserting SQL about all of them. Nine commits, all on `feat/non-sql-datasources`.

**Why.** The capability descriptor (`Shape` / `QueryLanguage`) landed earlier on this branch and nothing was reading it here. Meanwhile the prompt and the tools asserted SQL in a dozen places across both model paths. Handed that about a source with no tables, a model does not stop and ask — it resolves the contradiction in favour of the most concrete instruction (the tool description, attached to the argument it governs), writes SQL, and spends the turn on a query the source was always going to reject. This is item 1 of decisionbox-io/decisionbox-enterprise#254, one of the three things keeping `GA4_DATASOURCE_ENABLED` off.

**How.** Branch, don't generalise. Each SQL statement resolves on the turn's declared shape, and the SQL side of every branch is the original text unchanged. Three things turned out to matter more than the wording:

- **The set the branch reads.** Not the visible list: the router narrows what is *shown*, but `resolveQueryDatasource` validates a model-chosen id against the whole project on purpose and `search_tables` spans every datasource, so a cube the router passed over is one search result away. `reachable()` is the pin alone on a pinned turn and the whole project otherwise — and a passed-over cube is named at the end of the catalog, marked as not selected, since otherwise the model could be sent to a source whose language it had never seen.
- **Two facts, not one.** `anyCube` decides whether a blanket "SQL" statement is still *true*; `allCube` decides whether a table-shaped tool can work *at all*. `lookup_schema` is withheld only on the second — on a mixed turn it is still right for the SQL datasource — and that matters because while the turn is ungrounded the model is *forced* to call something, so an advertised-but-impossible tool can eat the step meant to gather evidence.
- **The correction outranks the prompt.** The grounding nudge arrives after the model has gone wrong, as the most recent thing in the conversation, and on the tool path as an error on the call it is correcting. Sending a cube turn back to `INFORMATION_SCHEMA` from there overrides everything the system prompt said, at the moment it is most likely to comply.

**Verified.**

- `make build`, `make lint-go` (all modules), `make lint-docs` clean. **CI green** — Go Build, Go Test, Go Lint, Docs Lint, License Check all pass on the head commit.
- `make test-go` passes everywhere except `libs/go-common/mongodb TestNewClient_UnreachableHost`, a **pre-existing environment artefact**: it expects `localhost:27099` to be unreachable and something in this container listens on it. That package is untouched by this branch and it fails identically on the base commit. CI does not hit it.
- **Byte-identity, measured not asserted.** Eight surfaces rendered under the base-branch implementation and under this one, compared byte-for-byte, re-run after every review round: `DIFFERING: NONE`. The grounding nudge is pinned the same way against its historic literal. An all-SQL project — every project that exists today — is unchanged.
- **46 distinct mutants, 45 killed.** Three survived first time. Two of them are the reason two tests exist (the loop's own tool wiring; `hiddenCubes` surfacing SQL datasources) and were killed after. The third — stripping the project list off the routing the prompt is built from — genuinely changes nothing observable, since `reachable()` falls back to the visible set and the two coincide on a single-datasource turn; that wiring is covered by `TestResolveTurnRouting_CarriesTheWholeProject` instead.
- **Codex: 8 rounds, 7 with findings, then clean.** In order: the reachable-vs-visible set, `search_tables`'s description, `lookup_schema` on the tool path, then on the text path, then the prompt still naming it, the grounding nudge, the tenant-scope wording, and the hidden reachable cube. Every one was a real contradiction between what the model was told and what it was handed — not style. Two are worth calling out: the reachable-set finding invalidated my first framing of the change, and the hidden-cube finding closed a residual I had written into the PR as deferred. No Copilot review was posted (auto-review is not active on this repo).

- **Nothing tells the model a non-SQL source's request format.** The prompt names the language and stops, deliberately: `search_tables` returns metrics and dimensions, not a request envelope, and describing a format the prompt cannot show would be worse than naming the language. So the first request against such a source is a guess — recoverable in principle via the adapter's repair prompt, except the fix-response extractor still requires `SELECT`. Written up with both options on decisionbox-io/decisionbox-enterprise#254; this is what still keeps the GA4 datasource gated off, together with that extractor.
- **Tenant-scope enforcement is unchanged and was not touched here.** `verifyFilter` is a substring check on query text for SQL and non-SQL alike: it catches a request that never mentions the scope field and cannot verify what one that does actually asks for. That is a pre-existing property, not something this change introduces — real scoping rests on the scoped credential and the governance middleware. A structured-*payload* query is refused outright rather than checked. I have deliberately left it alone: it is security-sensitive and belongs in its own change, not in a prompt PR. @abacigil worth a look if you think the substring check should be doing more.
- `summarize.go`'s over-fetch note still says "use aggregate SQL" — it fires only when a result exceeds the fetch cap, which a cube adapter caps for itself, so it is out of scope here rather than overlooked.
- No labels, milestone or assignee set, matching #363/#364/#365 on this branch; project board untouched per your earlier call.

— Co-coded with Jale 🤖

PR #367, opened 2026-09-02
feat(agent): repair a failed query in the language it was written in

jale-dbox:
## Why

The repair loop — the thing that turns a failed query into a corrected one — was SQL-only end to end. A source whose queries are a **request format** rather than SQL therefore had **no repair path at all**: its first error was final.

Two independent halves broke it, and each would have been enough on its own.

**1. The instruction demanded SQL.** The user message carrying the failed query is built here:

```
Fix this SQL query (attempt 1). Return ONLY the corrected SQL.

Query:
```sql
{"metrics":["nope"],"dimensions":["country"]}
```
```

That arrives *directly beneath* the source's own repair template, which opens `This source is NOT a database … it accepts no SQL`. The instruction is later and more concrete, which is the side a model takes — the same failure mode #366 spent its rounds on, one layer down.

**2. The extractor discarded the reply.** `extractFixedSQL` accepted a corrected query only if it contained `SELECT`, in every path: fenced block, raw text, and the `{"fixed_sql": …}` envelope. A corrected report request satisfies none of them, so it was thrown away and the attempt recorded as a parse failure — which reads, in the fix history, exactly like the loop ran and found nothing rather than like it could not run at all.

## What changed

**The instruction names the language it is asking for.**

```
Fix this query (attempt 1). It is written in Report Request (JSON), not SQL.
Return ONLY the corrected query.
```

No `sql` fence tag either — a tag names a syntax highlighter, and guessing one for an arbitrary request format puts a second, wrong claim about the language next to the right one. A warehouse's instruction is unchanged **character for character**, pinned against its historic literal.

**The extractor accepts a structured request** — a JSON object that *is* the query — alongside SQL, whether it arrives raw, fenced, or wrapped. The SQL paths keep precedence and are untouched.

The `{"fixed_sql": …}` envelope needed generalising too, which review caught and I had missed. Models wrap their answer in it whether or not the prompt asked for a wrapper — that is why the path exists — and a model repairing a request-format source wraps a **request**, not a string of SQL. The old decode used a struct with a string field, which fails outright on an object value, so the whole envelope was discarded. It is now read three ways: a string containing `SELECT`, a string that is itself a request, and an object that is one.

The wrapper is still **never handed on as the query**. If its contents are not one, the reply carried none — and submitting the model's packaging to the source would be worse than failing, since it parses as JSON and nothing downstream would have stopped it.

**Template and language are set together, by one constructor** (`ai.NewQueryFixerFor`), read through the `warehouse.QueryRunner` seam rather than off the SQL surface. The pairing is the point: a fixer given the source's own repair template but the SQL instruction is *worse* than one given neither, because the model then reads "this source accepts no SQL" immediately above an instruction demanding it. Three call sites (ask-serve, single-warehouse discovery, multi-warehouse discovery) now go through it.

**Is it SQL? Answered from a declaration, not a dialect label.** `warehouse.NonSQLLanguage(p)` returns `""` for a warehouse and the source's own language otherwise, on the basis that a provider implements `QueryRunner` *precisely because* its queries are not SQL — a SQL provider gains nothing by implementing a seam the adapter already maps onto its SQL surface. The alternative was substring-matching `"SQL"` against labels like `PostgreSQL` / `T-SQL` / `BigQuery Standard SQL`, which happens to work today and would break the first time someone registers a dialect without those three letters.

## What is unchanged for warehouses

One behaviour differs, and it is worth stating rather than burying: a reply that is a **bare JSON object and not a `fixed_sql` envelope** used to be refused by the extractor and is now handed on, where the warehouse rejects it as a syntax error. Both outcomes are a failed attempt inside the same retry loop; only the recorded error differs. Everything else — the instruction text, the `SELECT` acceptance, the envelope handling, the error paths — is byte-identical.

## Verification

- `make build`, `make test-go`, `make lint-go` (all modules), `make lint-docs` — clean. The one pre-existing failure, `libs/go-common/mongodb TestNewClient_UnreachableHost`, is an environment artefact: it expects `localhost:27099` to be unreachable and something in this container listens on it. Untouched by this branch, fails identically on the base commit, and does not occur in CI.
- New tests in `internal/ai/query_fixer_language_test.go` and `libs/go-common/warehouse/capability_test.go`:
  - **`TestFixInstruction_SQLIsUnchanged`** — the warehouse instruction against its exact historic literal.
  - **`TestFixInstruction_NamesTheLanguageItIsAskingFor`** — no `sql` fence, no "corrected SQL", names the language.
  - **`TestExtractFixedSQL_AcceptsAStructuredQuery`** — raw, `json`-tagged and untagged fences.
  - **`TestExtractFixedSQL_StructuredAcceptanceDoesNotSwallowAnEnvelope`** — the envelope still yields its contents; an envelope holding no query is refused rather than returned whole; `{}` is refused.
  - **`TestExtractFixedSQL_ReadsAStructuredQueryOutOfAnEnvelope`** — the request carried as an object, as a stringified object, and SQL in the envelope unchanged; plus four envelope values that are not queries (`null`, a number, `{}`, prose) which must all still be refused.
  - **`TestOrchestrator_NewQueryFixer_ReadsTheSourceNotTheSQLSurface`** and **`TestAskQueryFixer_ReadsTheSourceNotTheSQLSurface`** — the template and the instruction arrive as a matched pair at both construction sites.
  - **`TestExtractFixedSQL_StillRefusesProse`** — including a JSON *array*, which the new path must not accept.
  - **`TestSQLFixer_RepairsANonSQLQueryEndToEnd`** — the whole fixer through a mock provider, because the instruction and the extractor are separately correct and still leave the source stranded if they are wired to different languages.
  - **`TestNewQueryFixerFor_TakesBothFromTheSource`** and **`TestNonSQLLanguage_AnswersFromTheDeclarationNotADialectLabel`** — the wiring and the signal.
- **19 mutations, 17 killed.** The two survivors are **equivalent mutants**, not coverage gaps, and both are now commented as such: removing the leading-brace fast path changes nothing (decoding into a `map[string]json.RawMessage` already rejects an array, a number, a string and `null`), and dropping the explicit present-key check changes nothing (a missing key yields a nil `RawMessage` that every branch below rejects anyway). Killed along the way: both new helpers reverting to the SQL surface, the envelope dropping either structured shape, and the envelope accepting an object that is not a query.
- **Patch coverage.** The first push came in at 67% — below every earlier PR in this series — because the three sites that build a fixer sit inside runtime builders and orchestration methods with no test harness. Each now goes through a small named function (`newQueryFixer` on the orchestrator, shared by both discovery paths, which were duplicating it anyway; `askQueryFixer` for an Ask turn), asserted through the **rendered dialog** rather than through struct fields — a fixer configured correctly and still rendering "Fix this SQL query" at a source that accepts none is the failure that matters, and reading the struct back would not catch it. All three helpers and every new function in `sql_fixer.go` are at 100%.

## Follow-up

This does not close everything that keeps a cube datasource gated off. The prompt still names a non-SQL source's *language* without showing its *request format*, so the model's **first** query is a guess — now recoverable in one repair round instead of failing outright, which is the difference this PR makes, but still a wasted round-trip on every first query. Surfacing the format up front needs a descriptor change across both repos; tracked with its trade-offs on decisionbox-io/decisionbox-enterprise#254.

— Co-coded with Jale 🤖


jale-dbox:
## Final report

**What.** The repair loop now works in whatever language a source's queries are written in. Four commits.

**Why.** A source whose queries are a request format rather than SQL had **no repair path at all** — its first error was final. Two independent halves broke it: the instruction carrying the failed query said "fix this SQL query" inside a ```sql fence, directly contradicting the source's own repair template printed above it; and any reply that did come back was discarded, because the extractor required `SELECT`. The attempt was recorded as a parse failure, so the loop looked like it had run and found nothing rather than like it could not run at all. This is one of the two remaining items on decisionbox-io/decisionbox-enterprise#254.

**How.** The instruction names the language and drops the `sql` fence tag; the extractor accepts a structured request raw, fenced, or wrapped in a `{"fixed_sql": …}` envelope. Template and language are read through the `QueryRunner` seam and set **together** by one constructor — a fixer given the source's own template but the SQL instruction is worse than one given neither. Whether a source's queries are SQL is answered from a declaration (a provider implements `QueryRunner` *because* its queries are not SQL), not by substring-matching `"SQL"` against dialect labels like `PostgreSQL` / `T-SQL`, which happens to work today and breaks on the first label without those three letters.

**Verified.**
- **CI green** on the head commit — Go Build, Go Test, Go Lint, Docs Lint, License Check, and codecov/patch.
- `make test-go` passes everywhere except `libs/go-common/mongodb TestNewClient_UnreachableHost`, a pre-existing environment artefact (it expects `localhost:27099` to be unreachable; something in this container listens on it). Untouched by this branch, fails identically on the base commit, does not occur in CI.
- **19 mutations, 17 killed.** Both survivors are equivalent mutants, now commented as such in the code: the leading-brace fast path and the explicit present-key check are each redundant with a decode that already rejects the same inputs.
- The warehouse instruction is pinned against its exact historic literal, so a SQL repair is character-for-character what it was.
- **Codex: 3 rounds** — clean, then one P2, then clean.

**Two things worth flagging in review.**

1. **Review caught a real gap I had missed.** I excluded the `{"fixed_sql": …}` envelope from the structured path on the reasoning that its contents are the query. That was right about the wrapper and wrong about the contents: a model repairing a request-format source wraps a *request*, and the old decode (a struct with a string field) failed outright on an object value, so the whole envelope was discarded. Now read three ways. My original PR description argued for the exclusion; it has been corrected rather than quietly edited.

2. **Patch coverage failed on the first push**, at 67% — below every earlier PR in this series. The uncovered lines were the three sites that build a fixer, each inside a runtime builder or orchestration method with no test harness. Rather than wave it through I extracted each into a small named function and asserted it through the **rendered dialog** — a fixer configured correctly but still rendering "Fix this SQL query" at a source that accepts none is the failure that matters, and reading struct fields back would not catch it. The orchestrator pair was duplicated anyway. All three helpers are now at 100%.

**Known limitation.** This makes a wrong first query *recoverable*; it does not make it *unlikely*. Nothing shows the model a non-SQL source's request format — the prompt names the language and stops, because `search_tables` returns metrics and dimensions, not a request envelope. So the first query against such a source is still a guess that costs one repair round. Surfacing the format up front needs a descriptor change across both repos; tracked with its trade-offs on decisionbox-io/decisionbox-enterprise#254.

— Co-coded with Jale 🤖

PR #368, opened 2026-09-02
feat(api): refuse a project with no data source that can carry it

jale-dbox:
## Why

A data source can declare that it is only worth having *alongside* something else — an observation source rather than a system of record. That declaration was already stored, already resolved through `EffectiveAnchoring`, and already used to refuse an illegal promotion. It was read by nothing that could act on it.

So a project holding **nothing but** such sources was configurable, and it failed in the worst available way: not with an error, but with **analysis**. It reached the agent looking perfectly healthy, ran to completion, and produced a confident restatement of what those sources' own reporting already shows for free. Nothing in the output points back at the configuration — which is exactly why this belongs at the point of configuration, where the message can still name the fix.

## What changed

**`models.AnyAnchors(whs)`** — does this set contain anything that can carry a project? It resolves through `warehouse.EffectiveAnchoring`, so the provider's declaration is the ceiling and a per-source override may only demote. That matters for the case the naive version misses: a source whose provider *could* anchor, demoted by hand, leaving the project with nothing.

Refused in three places, which is two more than it looks:

| where | why it needs its own check |
|---|---|
| project create | the obvious one |
| settings edit (`warehouse` swap) | reaches the same end state one edit later, through a different route |
| discovery run | a project can arrive in this state without passing either route — it may predate the rule, or its provider's declaration may change between releases |

**An empty set passes everywhere.** A project with no data sources yet is how every project starts; refusing it here would make the product unusable in order to say something true. The run paths refuse an empty set on their own terms.

**An unregistered provider still resolves to anchoring.** A binary that has not linked a provider cannot construct the datasource either, so that project is broken in a more obvious way, earlier — making anchoring depend on which providers a binary happens to blank-import would be invisible at the call site.

## Verification

- `make build`, `make test-go`, `make lint-go` (all modules), `make lint-docs` — clean. The one pre-existing failure, `libs/go-common/mongodb TestNewClient_UnreachableHost`, is an environment artefact: it expects `localhost:27099` to be unreachable and something in this container listens on it. Untouched by this branch and failing identically on the base commit; it does not occur in CI.
- New tests, built on the existing `anchoringProbe` fixture:
  - **`TestRejectUnanchoredProject`** — seven sets, including the two that a naive implementation gets wrong: *the only source was demoted by hand* (must refuse) and *an unregistered provider* (must not refuse). Also asserts the refusal names the fix, since a 400 that doesn't is not much better than the silence it replaces.
  - **`TestProjectRoutes_RefuseLeavingAProjectWithNothingToCarryIt`** — both HTTP routes, and that a refused datasource is not persisted anyway.
  - **`TestDiscoveriesHandler_TriggerDiscovery_RefusesAProjectNothingCanAnchor`** — the run-path refusal, and that an anchor alongside enrichment sources is enough.
- **7 mutations, all killed.** Two survived on the first pass and are the reason the HTTP-level tests exist: the helper was covered and neither call site was, so the refusal could be deleted from both routes without a single failure. Also killed: `AnyAnchors` reading the override directly instead of the resolved value, and inverting *any* into *every*.

## Scope

This is the platform half of the anchoring work. The enterprise data-source management routes (add / edit / remove) mutate the same set and need the same invariant — a follow-up on that side, since removing the last system-of-record source is the most likely way to reach this state in practice.

— Co-coded with Jale 🤖


jale-dbox:
## Final report

**What.** A project must now have at least one data source that can carry an analysis. One commit.

**Why.** A source can declare it is only worth having alongside something else. That declaration was stored, resolved, and used to refuse an illegal promotion — and then read by nothing that could act on it. A project holding nothing but such sources failed in the worst available way: not with an error, but with **analysis**. It reached the agent looking healthy and produced a confident restatement of what those sources' own reporting already shows. This is the enforcement half of W6 on decisionbox-io/decisionbox-enterprise#252, and it closes the gap at `projects.go:737` that was accepted as scoped-out during #362.

**How.** `models.AnyAnchors` resolves through `warehouse.EffectiveAnchoring`, so a provider's declaration is the ceiling and a per-source override may only demote. Refused on project create, on the settings edit that swaps a single-warehouse project's only datasource, and again on the discovery run path — a project can arrive in that state without passing through either route, because it may predate the rule or its provider's declaration may change between releases. An empty set passes everywhere: a project with no data sources yet is how every project starts.

**Verified.**
- **CI green** on the head commit — Go Build, Go Test, Go Lint, Docs Lint, License Check, codecov/patch.
- `make test-go` passes except the pre-existing `libs/go-common/mongodb TestNewClient_UnreachableHost` environment artefact (expects `localhost:27099` unreachable; something here listens on it) — untouched by this branch and failing identically on the base commit.
- **7 mutations, all killed.** Two survived first, and they are why the HTTP-level tests exist: the helper was covered and neither route was, so the refusal could be deleted from both call sites without a single failure. Also killed: reading the override directly instead of the resolved value, and inverting *any* into *every*.
- Tests cover the two cases a naive implementation gets wrong — the only source demoted by hand (must refuse) and an unregistered provider (must not) — and assert the 400 names the fix, since a refusal that doesn't is not much better than the silence it replaces.
- **Codex: 1 round, clean.**

**Follow-up, not done here.** The enterprise data-source routes (add / edit / remove) mutate the same set and need the same invariant. Removing the last system-of-record source is the most likely way to reach this state in practice, and that route lives in the enterprise repo — it is blocked on this PR merging, since it needs `models.AnyAnchors`.

— Co-coded with Jale 🤖

PR #369, opened 2026-09-02
refactor(warehouse): share the dataset requirement with everything that configures a source

jale-dbox:
## Why

*"Does this source need a dataset configured?"* is asked wherever a datasource is configured — by the agent when it builds a provider, and by whatever validates the request that created it in the first place. It lived **unexported inside the agent**, so the second caller has no way to ask and has to guess.

Guessing has a wrong answer in each direction, and both are quiet:

- **Demand one from a source that has none** → the source is unconfigurable, or an operator invents a value that means nothing and then wonders why it is ignored.
- **Drop the demand** → a real warehouse is configurable with nothing to query, and it surfaces much later as an empty index.

Two copies of this rule would be two chances to disagree about whether a source can be configured at all.

## What changed

`requiresDataset` moves from `agentserver` to **`warehouse.RequiresDataset`**, beside the capability descriptor it reads. No behaviour change: same registry lookup, same defaults, and the agent's own call site now goes through it.

The default is the part worth keeping an eye on — an **unregistered slug still requires a dataset**. That is what every provider needed before shape existed, and the opposite default would silently drop the check for a real warehouse whose provider a binary happens not to link.

The test moves with the function. Its stub provider registrations stay in the agent package but in **a file of their own**: the registry is process-global and a second test there (`TestProjectNeedsBlurbs`) reads the same slugs, so a package whose stubs live in whichever test file first needed them breaks the moment that file moves — which is exactly what happened while writing this, and is why the file exists.

## Verification

- `make build`, `make test-go`, `make lint-go`, `make lint-docs` — clean. The one pre-existing failure, `libs/go-common/mongodb TestNewClient_UnreachableHost`, is an environment artefact (it expects `localhost:27099` to be unreachable; something in this container listens on it), untouched by this branch and failing identically on the base commit.
- `TestRequiresDataset` moves to the warehouse package, covering the cube case, the table case, and both flavours of unregistered slug.
- **New: `TestInitWarehouseProvider_DatasetRequirementFollowsShape`** — the call site had no coverage at all, so the requirement could be deleted from it, or applied to every source regardless of shape, without a single failure. Both of those are now caught.
- **5 mutations, all killed**, including: an unregistered provider no longer needing a dataset, a cube source being asked for one, and the two call-site mutants the new test exists for.

## Scope

This unblocks a follow-up in the enterprise data-source routes, which reject any datasource configured without a dataset and so cannot currently accept a source that has none.

— Co-coded with Jale 🤖

### DIFF
--- a/libs/go-common/warehouse/shape.go
+++ b/libs/go-common/warehouse/shape.go
@@ -1,0 +1,26 @@
+package warehouse
+
+// RequiresDataset reports whether a provider needs a dataset configured.
+//
+// Everything table-shaped does — a dataset is where its tables live. A
+// cube-shaped source has none, so demanding one makes it unconfigurable, or
+// forces an operator to invent a value that means nothing and then wonder why
+// it is ignored.
+//
+// An unregistered slug requires one. That default matters as much as the cube
+// case: it is what every provider needed before shape existed, and the
+// opposite default would silently drop the check for a real warehouse whose
+// provider a binary happens not to link.
+//
+// This lives beside the capability descriptor rather than at a call site
+// because the question is asked wherever a datasource is configured — by the
+// agent when it builds a provider, and by whatever validates the request that
+// created it. Two copies would be two chances to disagree about whether a
+// source is configurable at all.
+func RequiresDataset(providerSlug string) bool {
+	meta, ok := GetProviderMeta(providerSlug)
+	if !ok {
+		return true
+	}
+	return meta.EffectiveShape() != ShapeCube
+}
--- a/libs/go-common/warehouse/shape_test.go
+++ b/libs/go-common/warehouse/shape_test.go
@@ -1,0 +1,47 @@
+package warehouse
+
+import (
+	"testing"
+)
+
+// shapeStubFactory satisfies the registry's factory signature without doing
+// anything — registration only needs a non-nil factory.
+func shapeStubFactory(ProviderConfig) (Provider, error) { return nil, nil }
+
+func init() {
+	RegisterWithMeta("shape-test-tabular", shapeStubFactory, ProviderMeta{
+		Name: "Tabular test source",
+	})
+	RegisterWithMeta("shape-test-cube", shapeStubFactory, ProviderMeta{
+		Name:       "Cube test source",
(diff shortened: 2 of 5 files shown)


jale-dbox:
## Final report

**What.** `requiresDataset` moves out of the agent and into `warehouse.RequiresDataset`. One commit, no behaviour change.

**Why.** "Does this source need a dataset configured?" is asked wherever a datasource is configured — by the agent when it builds a provider, and by whatever validates the request that created it. It lived unexported inside the agent, so the second caller has to guess, and guessing is quietly wrong in both directions: demand one from a source that has none and it becomes unconfigurable; drop the demand and a real warehouse is configurable with nothing to query, surfacing much later as an empty index.

**How.** Same registry lookup, same defaults, now beside the capability descriptor it reads. The agent's own call site goes through it. The unregistered-slug default still requires a dataset — that is what every provider needed before shape existed, and the opposite default would silently drop the check for a real warehouse whose provider a binary happens not to link.

**Verified.**
- **CI green** on the head commit — Go Build, Go Test, Go Lint, Docs Lint, License Check, codecov/patch.
- `make test-go` passes except the pre-existing `libs/go-common/mongodb TestNewClient_UnreachableHost` environment artefact, untouched by this branch and failing identically on the base commit.
- **5 mutations, all killed** — including an unregistered provider no longer needing a dataset, and a cube source being asked for one.
- **New: `TestInitWarehouseProvider_DatasetRequirementFollowsShape`.** The call site had no coverage at all, so the requirement could be deleted from it *or* applied to every source regardless of shape without a single failure. Both are now caught.
- **Codex: 1 round, clean.**

**One thing worth knowing about the move.** Deleting the old test file broke `TestProjectNeedsBlurbs`, which was silently relying on the stub provider registrations in that file's `init()`. The registry is process-global, so a package whose stubs live in whichever test file first needed them breaks the moment that file moves. They now live in a file of their own, which is the actual fix rather than putting the deleted file back.

**Why this exists.** It unblocks the enterprise data-source routes, which currently reject any datasource configured without a dataset and so cannot accept a source that has none.

— Co-coded with Jale 🤖

Issue #252, opened 2026-05-28
Show insight validation status in all insight lists (Insights tab, discovery results, search)


erenbaydemir:
## Problem

An insight's validation result (confirmed / supported / rejected / partial / unverifiable / …) is currently only visible after opening a single insight's detail view. When scanning a **list** of insights, there is no way to tell which insights were validated, rejected, or never verified.

This makes it hard to judge at a glance which insights to trust when looking at:
- the **Insights tab** (all insights for a project),
- a **discovery run's results**, and
- **search results** (including cross-project search).

The user has to open each insight one by one to see its validation verdict.

## Proposed Solution

Show each insight's validation status inline, in every place where insights are listed, using the existing badge component.

- Display the validation verdict as a small badge/column in each insight row/card on:
  - the Insights tab,
  - discovery run results,
  - search results.
- Use `Insight.Validation.Combined` as the canonical status (the new-shape verdict). Possible values: `confirmed`, `supported`, `rejected`, `partial`, `unverifiable`, `validation_disabled`, `skipped_budget_cap`.
- Reuse the existing UI so colors/labels stay consistent:
  - `ui/dashboard/src/components/validation/VerdictBadge.tsx`
  - `ui/dashboard/src/components/validation/statusMeta.ts` (labels, taglines, colors)
- Handle the "no validation" case (validation not run or disabled) with a clear neutral state, so the column is consistent across rows.
- Nice to have: let users sort/filter the Insights tab by validation status.

## Implementation pointers

Validation status is already in most list payloads, so most of the work is frontend:

- Status source (Go): `Insight.Validation` → `Combined` field. Status constants in `libs/go-common/models/validation/status.go`.
- Already returns `validation` in the response: insights list (`services/api/internal/handler/insights.go`), discovery results (`services/api/internal/handler/discoveries.go`), single-insight detail.
- **Needs a small API change — search omits it:** `searchResultItem` in `services/api/internal/handler/search.go` does not include validation. Add the status there, and add the field to `SearchResultItem` in `ui/dashboard/src/lib/api.ts`.
- Frontend list surfaces to update (add the badge / column):
  - Insights tab: `ui/dashboard/src/app/projects/[id]/insights/page.tsx` (the rendered table, and the semantic-search results table on the same page).
  - Discovery run results: `ui/dashboard/src/app/projects/[id]/discoveries/[runId]/page.tsx` (the `InsightRow` component).
[...]

PR #376, opened 2026-09-02
feat(askserve): let a turn declare what it joined two datasources on (#252 W5)

jale-dbox:
Part of #252 (W5 — join-key validation). Second half of the direction on cross-datasource hops; the first half shipped in #375.

## The problem

A turn combines datasources in hops: query one, then filter another by values read out of that result. Whether the hop is sound turns on something no amount of looking at the query can settle — that the field those values came from names the same entities on both sides. An `order_id` that is an order id in the warehouse and a session id in the analytics source produces a perfectly well-formed answer about the wrong rows, and nothing in the result says so.

## What this adds

**`joins_on: {source_step, field}`** — an optional argument on `query_data`, offered only on a multi-datasource turn, where the model says which earlier step it took the values from and which column they came from. `field` is the **source-side** name deliberately: that is the one the model actually observed, so the claim can be checked against what that step returned.

**A result now says whether its scope was verified.** `QuerySummary.Scoped` is set only on a query made after a *different* datasource was queried in the same turn — nil, and therefore absent from the persisted summary and the observation, on every single-datasource turn and on the first hop of a multi one.

| | outcome |
|---|---|
| declared, and a join-key report confirms the field | `scoped: true`, nothing added to the observation |
| declared, but no report / the report lists other keys / the report is unreachable | `scoped: false` + a note saying which |
| **not** declared, on a turn that already queried another datasource | `scoped: false` + a note naming `joins_on` |
| declared, but self-contradictory | the query is **rejected** before it runs |

**Omission never blocks** — that was the explicit constraint, and it is what keeps the existing SQL multi-hop path working unchanged. A *self-contradictory* declaration is a different thing and is rejected: naming a step that never ran, a step on this same datasource, or a column that step never returned is a false claim the turn can check, and the rejection hands back what is needed to repair it (the real step ids, the real columns). It costs no warehouse work — the verdict resolves before the connection is acquired.

**`libs/go-common/agentplugin`** gains the seam the connect-time binding report plugs into. It has to live outside `internal/`, and it is deliberately one-sided: `Verified: true` requires a report to positively name the field; an absent report, an unindexed datasource, and a validator that errors all land on false, because none of them proves the join is *wrong* either. With nothing wired — a platform-only deployment — every declared join comes back unverified rather than fine.

**Telemetry** (`cross_datasource_query`, coarse and id-free: `declared` + `outcome`) is the evidence for the follow-up already scheduled — tightening an undeclared cross-datasource filter from a caveat to a refusal, once the numbers show the model populates the declaration reliably. It carries no field name, datasource id, or project.

## Verification

- `make build`, `make lint-go` (0 issues), `make lint-docs`, both module suites, `-count=2` on the touched packages. CI green; `MERGEABLE / CLEAN`.
- **Mutation-tested throughout, six batches.** It found four things the reviewer did not: a hardcoded step round that every test accepted, an unreachable guard, an untested cube branch, and a line that provably cannot change behaviour (deleted rather than kept as defence). Two survivors are genuine **equivalent mutants**: swapping `hasJSONPayload` for a length check on `null` reaches the same answer by a different route, and clearing the verdict inside the panic recover is unobservable because a panicking call never assigned it. Recorded as equivalent rather than counted as covered.
- Three known `-count=2` failures in `libs/go-common/llm` and one in `libs/go-common/mongodb` (a local port collision) are pre-existing on the bridge — verified unmodified there.

## Review history — seven Codex rounds at `xhigh`, the last one clean

Five P2s and a P3, every one valid and fixed:

| # | Finding | |
|---|---|---|
| 1 | Same-batch tool calls read as observed | Could stamp `scoped: true` on a hop that provably never happened, and caveat two independent queries that never touched each other |
| 2 | `scoped` claimed more than it verified | Narrowed the claim. The suggested remedy — check the query text for the observed values — is the bug #375 removed, so it was deliberately not built |
| 3 | A plugin panic could kill the agent | A turn goroutine has no recover; a nil map in a validator would take down every concurrent turn over a question whose honest answer was "cannot tell" |
| 4 | The text parser silently dropped malformed declarations | Filed real attempts as non-attempts in the telemetry meant to decide whether attempts can be relied on |
| 5 | The text prompt never showed the shape #4 made mandatory | A refusal the prompt gives no way to satisfy. Created by the previous fix |
| 6 | A leftover envelope stub outranked a real declaration | #4's failure reached another way; separated precedence from validity |

**Round 7 is clean.** It ran to completion on `e584b9b` and reported no findings: *"I did not identify any discrete correctness issues introduced by this patch. The added join-key validation path appears covered by targeted tests and preserves existing single-datasource behavior."* An earlier attempt at round 7 had died on the Codex account's usage limit — worth noting because that failure exits 0 and prints a review-shaped tail, so an interrupted round reads exactly like a clean one unless you go looking for the error string. Before quota returned I also re-read the whole diff myself and found one thing: the file-level comment enumerated the rejections and had gone stale as rejections were added. Fixed here.

## Follow-up

The enterprise side wires `datasources/binding` into the seam and documents the hop in `docs/multi-warehouse.md`. Until it lands, a declared join is checked for provenance but not compatibility — and says so.

— Co-coded with Jale 🤖

### DIFF
--- a/libs/go-common/agentplugin/joinkey.go
+++ b/libs/go-common/agentplugin/joinkey.go
@@ -1,0 +1,117 @@
+package agentplugin
+
+import (
+	"context"
+	"fmt"
+	"sync"
+)
+
+// A cross-datasource answer is assembled in hops: query one datasource, then
+// filter another by values read out of that result. The hop is only sound if
+// the field those values came from identifies the same things in both
+// datasources — an order id that is an order id on both sides, not an order id
+// on one and a session id on the other. Nothing in the query text says which of
+// those it is, and the agent cannot tell them apart by looking.
+//
+// The knowledge of which fields bind two datasources is produced when a
+// datasource is connected, by a component that reads both sides' fields. That
+// component is not part of this repository, so this registry is the seam it
+// plugs into: with a validator wired, a declared join key is checked; with none
+// wired, the answer is "cannot tell", never "fine".
+
+// JoinKeyRequest asks whether Field, as observed on the source datasource,
+// binds that datasource to the target one.
+type JoinKeyRequest struct {
+	ProjectID string
+	// SourceDatasourceID is the datasource the values were read FROM.
+	SourceDatasourceID string
+	// TargetDatasourceID is the datasource now being filtered by those values.
+	TargetDatasourceID string
+	// Field is the field name as it appeared in the source result.
+	Field string
+}
+
+// JoinKeyVerdict is a validator's answer.
+//
+// Verified is deliberately one-sided: true means a report positively names this
+// field as a key binding the two datasources. False means only that nothing
+// established it — an absent report, an unindexed datasource and a report that
+// lists other keys all land here, and none of them proves the join is wrong.
+// Detail says which, in words a model can act on, and is what the turn shows
+// instead of a verdict it has not earned.
+type JoinKeyVerdict struct {
+	Verified bool
+	Detail   string
+}
+
+// JoinKeyValidatorFunc answers a JoinKeyRequest. An error means the validator
+// could not reach its evidence; callers must treat that as "cannot tell" and
+// carry on, never as a refusal — a report being unreachable is not a reason to
+// fail a query the user asked for.
+type JoinKeyValidatorFunc func(ctx context.Context, req JoinKeyRequest) (JoinKeyVerdict, error)
+
+var (
+	joinKeyValidatorMu   sync.RWMutex
+	joinKeyValidator     JoinKeyValidatorFunc
+	joinKeyValidatorName string
+)
+
+// RegisterJoinKeyValidator registers fn as THE join-key validator. Plugins call
+// this from init() with a blank import. Empty name or nil fn panics, as does a
+// second registration: unlike the filter registries, which compose in order,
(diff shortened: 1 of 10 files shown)


jale-dbox:
## Round 7: clean

Codex quota came back, so the loop got finished properly rather than left at six. Round 7 ran to completion on `e584b9b` at `xhigh` and reported no findings:

> I did not identify any discrete correctness issues introduced by this patch. The added join-key validation path appears covered by targeted tests and preserves existing single-datasource behavior.

That is seven rounds: five P2, one P3, all valid, all fixed, and a clean seventh. CI is green on every check. PR body updated — it previously said this branch had never had a clean round, which was true when I wrote it and isn't now.

One thing worth keeping: the earlier attempt at round 7 died on the usage limit, and that failure **exits 0** and prints a review-shaped tail. An interrupted round is indistinguishable from a clean one unless you grep for the error string. I nearly filed one as the other.

The enterprise half is in progress — wiring the connect-time binding report into `RegisterJoinKeyValidator` so a declared hop gets checked against real evidence instead of the seam's "cannot tell". It doesn't need this merged to build, only to go green in CI.

— Co-coded with Jale 🤖

PR #377, opened 2026-09-03
feat(agent): record which catalog items are dimensions (#252 W5)

jale-dbox:
The catalog cache kept only the refs a catalog source offers, so nothing downstream could tell a **measure** from something that **identifies a record**. This adds the distinction.

## Why it matters

The two are not separable by name. A GA4 custom metric and a custom event-scoped dimension are both `customEvent:<name>`, so a metric called `score` is indistinguishable from a dimension called `score`. Anything asking *"could this field bind two datasources?"* was therefore free to answer yes about a number — and joining on a measure is meaningless.

Found by a Codex review of the query-time join-key work (#252 W5), which reads this cache. It applies to the connect-time binding report too, which has been reading the same untyped list.

## What changed

- **`SaveCatalog` takes the items rather than their refs.** The kind is only knowable at the index run, and the indexer already had it — it was throwing it away one line before the call.
- **`dimension_refs` is written alongside `refs`.** `Refs` is untouched, so retrieval and the staleness authority keep seeing the whole catalog. Nothing that reads `refs` today changes behaviour.
- **`FindCatalogDimensions` reads them**, and returns `nil` for a row written before this existed. That is deliberately the same answer as "no row at all": a legacy row *knows nothing about kinds*, which is not the same as knowing there are no dimensions, and falling back to `refs` would hand back metrics as though the question had been answered. A caller sees the datasource as unindexed for this question until it is indexed again — the safe direction, and the only honest one. The absence rule lives in the accessor so no consumer has to rediscover it.
- **A compile-time assertion that the repository satisfies `CatalogCache`.** Every use of that interface is a type assertion, so an implementation that stops matching doesn't fail to build — it silently stops being a `CatalogCache`, and the index run then remembers nothing while reporting success. Changing `SaveCatalog`'s signature did exactly that to two test fakes, and the unit suite caught it as two unrelated-looking failures. The assertion turns that into a build error.

## Verification

- `make build`, `make test-go`, `make lint-go` — clean. (`libs/go-common/mongodb`'s `TestNewClient_UnreachableHost` fails locally on a port collision; it is unrelated and fails the same way on the base branch.)
- Three new integration tests against a real Mongo (`-tags integration`): dimensions kept apart from metrics while `refs` still holds all four items, a legacy row reporting no dimensions, and a metric-only catalog. Existing catalog-cache integration tests updated to the new signature and still passing.

## Review

Two Codex rounds at `xhigh`; the second clean.

R1 (P2) caught the dimensions being derived from the *indexable* subset, which is deduplicated by ref and keeps the first item it saw — so a source listing a metric before a dimension of the same name would have dropped that dimension, and the other order would have kept it. The answer depended on the order a provider happened to return its catalog in, which is not a fact about the data.

They now come from the whole catalog, and a ref the source gives **more than one kind is left out entirely** rather than resolved by position: the source is calling one name two things, and neither answer is safe to hand to something deciding whether a field identifies a record. (Codex suggested reading kinds pre-dedupe, which fixes the dropped dimension but makes the ambiguous case *more* permissive — the wrong direction for a check whose only dangerous failure is a false yes.)

That fix also moved the save into `saveCatalogCache`, because `buildCatalogIndex`'s success path needs a live vector store: an inline block there could only ever have been checked by reading it.

## Note on rollout

Existing catalog rows carry no `dimension_refs`, so consumers of the new accessor treat those datasources as not-yet-known until their next index run. That is intentional — see above — and affects nothing that reads `refs`.

Part of #252

— Co-coded with Jale 🤖

### DIFF
--- a/services/agent/internal/database/catalog_cache_integration_test.go
+++ b/services/agent/internal/database/catalog_cache_integration_test.go
@@ -6,8 +6,19 @@ import (
 	"context"
 	"testing"
 
+	gowarehouse "github.com/decisionbox-io/decisionbox/libs/go-common/warehouse"
 	"github.com/decisionbox-io/decisionbox/services/agent/internal/models"
 )
+
+// dims renders refs as dimension items, which is what these cases are about —
+// the kind matters only to the tests that name it.
+func dims(refs ...string) []gowarehouse.CatalogItem {
+	out := make([]gowarehouse.CatalogItem, 0, len(refs))
+	for _, r := range refs {
+		out = append(out, gowarehouse.CatalogItem{Ref: r, Kind: gowarehouse.ItemKindDimension})
+	}
+	return out
+}
 
 const catalogTestProject = "catalog-cache-proj"
 
@@ -21,7 +32,7 @@ func TestAgentInteg_CatalogCache_RoundTrip(t *testing.T) {
 	r := NewSchemaCacheRepository(db)
 
 	refs := []string{"sessions", "activeUsers", "sessionDefaultChannelGroup"}
-	if err := r.SaveCatalog(ctx, catalogTestProject, "ga", "hash-a", refs); err != nil {
+	if err := r.SaveCatalog(ctx, catalogTestProject, "ga", "hash-a", dims(refs...)); err != nil {
 		t.Fatalf("SaveCatalog: %v", err)
 	}
 
@@ -58,7 +69,7 @@ func TestAgentInteg_CatalogCache_NotReadAsTableSchemas(t *testing.T) {
 	ctx := context.Background()
 	r := NewSchemaCacheRepository(db)
 
-	if err := r.SaveCatalog(ctx, catalogTestProject, "ga", "hash-a", []string{"sessions"}); err != nil {
+	if err := r.SaveCatalog(ctx, catalogTestProject, "ga", "hash-a", dims("sessions")); err != nil {
 		t.Fatalf("SaveCatalog: %v", err)
 	}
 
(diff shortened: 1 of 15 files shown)


jale-dbox:
## Final report

**What.** The catalog cache now records which of a catalog source's items are dimensions, alongside the refs it already kept.

**Why.** The two are not separable by name. A GA4 custom metric and a custom event-scoped dimension are both `customEvent:<name>`, so a metric called `score` was indistinguishable from a dimension called `score` — and anything asking *"could this field bind two datasources?"* was free to answer yes about a measure. Found by a Codex round on the query-time join-key work (enterprise #263), but it applies to the connect-time binding report from #258 too, which has been reading the same untyped list since it merged. The kind was known at the index run and thrown away one line before the write.

**How.** `SaveCatalog` takes the refs and the dimension refs; `dimension_refs` is stored beside `refs`, which is untouched, so retrieval and the staleness authority keep seeing the whole catalog. `FindCatalogDimensions` reads them and returns nil for a row written before this existed — the same answer as no row at all, because such a row *knows nothing about kinds*, which is not the same as knowing there are none. The absence rule lives in the accessor so no consumer rediscovers it.

**How it was verified.** `make build`, `make test-go`, `make lint-go` clean, CI green on every check. Four new unit tests for the dimension rule and the save wiring, plus three integration tests against a real Mongo: dimensions kept apart from metrics while `refs` still holds everything, a legacy row reporting none, and a metric-only catalog.

**Review: two Codex rounds at `xhigh`, the second clean.** R1 (P2) caught the dimensions being derived from the *indexable* subset, which is deduplicated by ref and keeps the first item seen — so a source listing a metric before a dimension of the same name would have dropped that dimension, and the reverse order kept it. The answer depended on the order a provider happened to return its catalog in.

They now come from the whole catalog, and a ref the source gives more than one kind is left out entirely rather than resolved by position. Worth noting that Codex's suggested remedy — read kinds pre-dedupe — fixes the dropped dimension but makes the ambiguous case *more* permissive, which is the wrong direction for a check whose only dangerous failure is a false yes.

That fix also moved the save into `saveCatalogCache`: `buildCatalogIndex`'s success path needs a live vector store, so an inline block there could only ever have been checked by reading it.

- Existing catalog rows carry no `dimension_refs`, so consumers of the new accessor treat those datasources as not-yet-known until their next index run. Intentional, and it affects nothing that reads `refs`.
- There is still no integration test driving the catalog index path end to end — it needs a live vector store, and none existed before this PR either. The save is covered at `saveCatalogCache`; the rest of that path is not.
- **This should merge before enterprise #263**, which reads the field this writes.

Part of #252

— Co-coded with Jale 🤖

PR #382, opened 2026-09-03
feat(discovery): stop cube exploration on no new signal, not a step count (#252 W12)

jale-dbox:
Exploration on a cube-shaped source now ends when it stops finding things, rather than when it reaches a number. **#252 W12**, platform-only — nothing in the enterprise repo changes.

## Why the floor cannot be tuned for this

`--min-steps` rejects a "done" signal below a fixed count. It exists for a real reason: reasoning models were terminating after 2–18 steps. It works because on a warehouse a step is roughly a table or a join, so counting steps approximates counting coverage.

That approximation does not survive a cube. There are no tables to work through — a query is a choice of metrics and dimensions — and the number of available slices is combinatorial. Sixty trivial breakdowns and fifteen revealing ones satisfy any floor equally well. Raising it buys nothing but quota spend against a source metered per request; lowering it restores the early termination the floor exists to prevent. **No value of the number is right, because the number is not the problem.**

## The rule

On a run that can query a cube, the floor is replaced by what it was standing in for: whether exploration is still turning anything up.

Each executed step is scored against the steps already taken — the run-scoped vector index is *already* embedding every step for the analysis phase, so this measures something the run was computing anyway. While recent steps are still finding new ground a completion is refused exactly as the floor refused it; three consecutive steps that repeat earlier work end the exploration, at whatever step that is. **`--max-steps` is untouched and remains the runaway cap.**

Which rule a run gets is read from the **registered shape** of its datasources' providers and logged at exploration start. A run that can only reach tables is unchanged, character for character — which is the point, since that is every deployment today.

## Four things this gets wrong if built the obvious way

**Novelty is asked before the step is indexed.** Afterwards the step is its own nearest neighbour, every measurement is 1.0, and every run reads as repetitive. `Nearest` also skips the step being judged, so a caller that gets the order wrong degrades instead of inverting.

**An unmeasurable step is not a repetition.** No index, a failing index, and a step with nothing worth embedding all answer *cannot tell* — and three of those must not read as three repeats. That would end runs on a silent measurement failure, producing a full set of insights that were never looked for. They clear the streak rather than extending it.

**A run that can never measure hands the decision back to the floor.** Otherwise a degraded vector store turns every run into a maximum-length one against exactly the sources that are metered per request — the opposite of the problem being solved. One successful measurement is enough to prove the machinery is there and keep the rule armed.

**The rule refuses an early completion itself rather than deferring to the floor when short of evidence.** `--min-steps` defaults to `0`. Deferring would accept a completion signalled on step one — reintroducing, inside the rule meant to replace the floor, precisely the early termination the floor was added to prevent. This one is a test, not a comment: it is the bug the first draft of this had.

**`Nearest` is a required interface method, not an optional one found by assertion.** This seam is wrapped by the orchestrator's telemetry decorator, and a wrapper that forgot to forward an optional method would strip the whole rule with nothing to notice. The compiler now refuses to build that; the decorator forwards it explicitly.

## Verification

- `go build ./...`, `go test ./...`, `golangci-lint run ./...`, `make lint-docs` — clean. One pre-existing failure is unrelated and reproduces on a clean checkout of the base: `libs/go-common/mongodb.TestNewClient_UnreachableHost`, which asserts a host fails to resolve and passes only where DNS agrees.
- **20 new tests.** The rule's decision table is written as observation sequences (`"nrrnr"`, `"nnn???"`) so the cases that distinguish a marginal-value rule from a second counter are legible: a novel step breaking a streak, unmeasurable steps not counting as repeats, and the floor being ignored in *both* directions.
- **`Nearest` is tested against a real Qdrant** (testcontainer, `-tags integration`): an empty index, an identical earlier step, the self-exclusion guard, and a step with no text. Those four answers are exactly what the rule reads as judgeable or not, and getting any of them wrong ends runs early or never.
- **Mutation tested, 12/12 killed** — including the threshold boundary, unjudgeable-counts-as-repeat, both directions of the measurability fallback, the deferral bug above, and both halves of the shape check.

## Two things I decided, worth a second opinion

**The trigger is "the run can reach a cube", not "every datasource is a cube".** An all-cube run cannot exist — a non-anchoring source may not run standalone Discovery (W6) — so the strict reading would make this dead code. A mixed run is exactly where a model can spend steps on trivial cube slices to satisfy a scalar floor, so that is where the rule is needed.

**Both constants are named, not flags.** `0.92` similarity and 3 consecutive repeats are properties of how steps embed, not of a deployment, and the epic's live evaluation is what should calibrate them. If you would rather they were operator-tunable alongside `--min-steps`, say so and I will add the flags.

Part of #252

— Co-coded with Jale 🤖


jale-dbox:
## Final report

**What.** Exploration on a run that can query a cube now ends when it stops finding things, rather than when it reaches a step count. **#252 W12**, platform-only.

**How.** Each executed data query is scored against the steps already taken, using the per-run vector index the analysis phase was already building. While recent steps are still turning up new ground a completion is refused exactly as the floor refused it; three consecutive steps that repeat earlier work end the exploration, at whatever step that is. `--max-steps` is untouched and remains the runaway cap. Which rule a run gets comes from the registered shape of the datasources it was actually **wired** for, and is logged at exploration start.

**Why.** `--min-steps` works because on a warehouse a step is roughly a table or a join, so counting steps approximates counting coverage. A cube has no tables and a combinatorial number of slices, so sixty trivial breakdowns and fifteen revealing ones satisfy any floor equally well. Raising it buys quota spend against a source metered per request; lowering it restores the early termination it exists to prevent. No value of the number is right, because the number was never the thing being measured.

**Verified.**
- `go build ./...`, `go test ./...`, `golangci-lint run ./...`, `make lint-docs` — clean. CI green on Go Build / Go Test / Go Lint / Docs Lint / License Check.
- 25 tests. The rule's decision table is written as observation sequences (`"nrrnr"`, `"n?n?n?n?n"`) so the cases separating a marginal-value rule from a second counter stay legible.
- `Nearest` is tested against a **real Qdrant** — an empty index, an identical earlier step, self-exclusion, a text-less step, an errored neighbour, 25 failures deep, and per-datasource scoping.
- **Mutation tested: 8/8 killed** in a final pass against the current code (threshold boundary, unjudgeable-as-repeat, streak reset, failed-query judging, schema actions as subjects, arming without an indexer, and both halves of the shape check), plus 3/3 on the Qdrant filter and 3/3 on the silence counter earlier.
- **Codex: 9 rounds, 12 findings, all valid, all fixed.** Three were caused by fixes to the previous round's finding — see below.

**Two red checks, neither from this code.**
1. **Integration Tests** only ran because I added `run-integration-tests` while trying to get `internal/discovery` covered (see limitation 1). With `ci.yml` back to base it still failed, on `TestLiteLLMTLS_WithCA_Succeeds` in the agent root package — container-dependent, passing on `main`, and untouched by this diff, which changes nothing outside `internal/ai` and `internal/discovery`. The label is removed; the result is stale.
2. **codecov/patch** — the uncovered lines are precisely the Qdrant-backed code whose tests CI does not run.

**The part worth a human eye.** Rounds 4 through 7 were the same defect wearing different clothes: each fix to the degraded-index fallback moved the failure to the adjacent case. The cause was inferring a property of the machinery from a tally of what the machinery once said — an observation recorded while the index looked healthy stays on the books after that stops being true. Round 7's change stopped patching and asked the question directly, once, where the decision is made. The result is **smaller** than it was two rounds earlier: one predicate and one counter fewer, and each measurement free to say the simple true thing.

Nine clean-ish rounds is evidence the implementation is sound. It is not evidence the **design** is what you want, and two decisions are mine rather than the issue's:

- **The trigger is "the run can reach a cube", not "every datasource is a cube".** An all-cube run cannot exist — a non-anchoring source may not run standalone Discovery (W6) — so the strict reading is dead code. A mixed run is exactly where a model can spend steps on trivial cube slices to satisfy a scalar floor.
- **`0.92` similarity and 3 consecutive repeats are named constants, not flags.** They are properties of how steps embed, not of a deployment, and the epic's live evaluation is what should calibrate them. Say the word and they become flags alongside `--min-steps`.

**Follow-ups / known limitations.**
1. **The Qdrant tests run nowhere but a developer's machine.** `internal/discovery` is absent from CI's integration package list. Adding it was tried and reverted: a pre-existing test there fails on a Qdrant gRPC connection reset on a CI runner. Documented in `docs/concepts/discovery-lifecycle.md` with the command to run them. This is one of **three** integration lanes across the two repos that execute nowhere — worth its own audit issue.
2. **The threshold is uncalibrated against a real embedding model.** The integration harness embeds by hashing, which proves the plumbing and says nothing about where 0.92 sits. That calibration belongs to the epic's live evaluation.
3. A run whose queries all fail, or whose index is unusable, falls back to `--min-steps` — which the CLI defaults to `0`, though the API sends `floor(0.6 × max_steps)`. Degraded in the direction of today's behaviour rather than toward a maximum-length run against a metered source.

— Co-coded with Jale 🤖

jale-dbox:
**CI is green.** `MERGEABLE/CLEAN` on `cc0d9bc` — Go Build, Go Test, Go Lint, Docs Lint, License Check all pass, and Integration Tests correctly **skips**.

For the record, since the earlier red check was mine rather than the code's:

`Integration Tests` is opt-in via the `run-integration-tests` label and shows *skipping* on every ordinary PR. It ran here only because I added that label while trying to get `internal/discovery` into CI coverage, and it failed on **`TestLiteLLMTLS_WithCA_Succeeds`** — `services/agent/litellm_tls_integration_test.go`, untouched by this branch, last changed by #339, and passing on `main`. Removing the label did not clear the check, because this workflow triggers on `opened`/`synchronize`/`reopened` and not on `labeled`, so the result stayed frozen until the next push.

Rather than push an empty commit to clear a cosmetic check, the commit that cleared it fixes what the other red check was actually pointing at: **`nearestFilter` had no coverage outside the integration lane**, which is a lane CI does not run — so the clause deciding which earlier steps a step may be judged against was, as far as any gate could see, untested. It now has four unit tests pinning its intent without a container, mutation 3/3 (dropping either clause, or letting a datasource-less step match everything, all die). The store's semantics stay with the integration tests, where they belong.

Nothing else changed. The design questions and known limitations in the final report above stand.

— Co-coded with Jale 🤖

jale-dbox:
**Confirmed as a real regression, and fixed. Thank you — this was the right catch.**

I reproduced it exactly as described before changing anything: floor of 60, `feed("nrrr")`, accepted at **step 4**.

The premise the replacement rested on does not hold. I had reasoned that a step count is meaningless on a cube — true — and concluded the floor should therefore step aside on any run that can reach one. But **every discovery run has tables**: a source that cannot anchor may not run discovery on its own (W6), so a cube is only ever present *alongside* a warehouse. The floor is a coverage proxy for that warehouse and remains a good one; connecting an enrichment source beside it changes nothing about it. The all-cube run where the floor really would be meaningless **cannot exist today** — so the rule was trading a real floor for a hypothetical benefit.

Your third point was the tell I should have followed myself: if the run log cannot name the floor, it is because the floor never fired, and a floor that never fires on a project that configured one is not a design choice, it is a bug.

**What changed.** The floor gates first. The marginal-value rule applies *past* it, where it can only keep exploration going — refused while recent steps are still turning things up, accepted once they stop. So a cube-reaching run is **never shorter than the same run is today, only longer**, and a floor refusal names the floor again. That direction guarantee is now a test, alongside your scenario as a named regression test.

Two of the previous tests asserted the buggy contract — `NoveltyIgnoresTheFloorInBothDirections` was, in hindsight, the bug written down as an intention. Both rewritten. Mutation confirms the ordering is load-bearing: putting the floor check back after the novelty branch fails the suite.

**What this gives up, so it is on the record.** The epic's "sixty trivial slices and fifteen excellent ones satisfy any floor equally well" complaint is no longer addressed — a run that spins on trivial cube slices still satisfies its floor by spinning. Fixing that *without* reintroducing your regression means counting only non-repeating steps toward the floor, which makes such a run **longer** rather than shorter, and spends more quota on the sources that are metered per request. That is a product trade-off rather than a correctness one, so I have not made it unilaterally. Say the word if you want it and it is a small change.

— Co-coded with Jale 🤖

jale-dbox:
**Ready. Clean round on `1999e8f`, CI green, `MERGEABLE/CLEAN`.**

> No actionable correctness issues were found in the changed code. Targeted tests for the modified agent packages pass.

The round before it caught the tail of the floor regression: both flag reference tables still said `--min-steps` is *ignored* on a cube-reaching run. That stopped being true when the floor was restored to gate first, and an operator reading either would have misdiagnosed a rejected completion and tuned the wrong setting — the same failure as the regression itself, a floor nobody can see. Corrected in `1999e8f`; the concept doc and CHANGELOG already carried the fix.

**Final tally: 13 rounds, 15 findings, all valid, all fixed.**

Worth stating plainly for whoever reviews this: **none of those 15 was the important one.** Every round checked whether the code did what it said, and it did. What was wrong was what it said — the claim that a step count is meaningless on a cube-reaching run, which is false because every such run also has a warehouse. Two of my own tests encoded that claim as an intention, one of them named `NoveltyIgnoresTheFloorInBothDirections`. Reviewing against a wrong specification produces clean rounds indefinitely.

So the design is worth a read on its own terms, not on the strength of the round count:

- **The trigger** is "the run can reach a cube". Given the floor now gates first, this only decides where the *extension* applies, which makes it a much smaller decision than it was.
- **`0.92` similarity and 3 consecutive repeats** are named constants, not flags — properties of how steps embed rather than of a deployment. One-line change if you want them tunable.
- **The epic's "sixty trivial slices satisfy any floor" complaint is no longer addressed.** Fixing it without reintroducing the regression means counting only non-repeating steps toward the floor, which makes such runs *longer* and spends more quota on precisely the metered sources. A product trade-off; not mine to make.

Known limitation unchanged: the Qdrant-backed tests live in `internal/discovery`, which is absent from CI's integration package list, so they run only locally — `cd services/agent && go test -tags=integration ./internal/discovery/`. Documented in `docs/concepts/discovery-lifecycle.md`.

— Co-coded with Jale 🤖

Issue #383, opened 2026-09-04
Discovery's step floor counts steps, not coverage — sixty trivial slices satisfy it


jale-dbox:
## The problem

`--min-steps` is the only thing standing between a reasoning model and a
three-step discovery run, and it counts **steps**, not **coverage**.

On a warehouse the two are close enough: one step is roughly one table or one
join, the catalog is finite, and working through it is what the floor is
approximating. On a cube-shaped datasource they come apart completely. A query
there is a choice of metrics and dimensions — on the order of 200 × 150
candidate slices with no join graph to prune by — so **sixty trivial slices and
fifteen revealing ones satisfy a floor of sixty equally well**. Re-tuning the
number only moves where it bites: raising it spends more quota against a source
metered per request, lowering it restores the early termination the floor exists
to prevent.

## Why this is open rather than done

#382 set out to fix exactly this, by replacing the floor with a marginal-value
rule on cube-reaching runs: keep exploring while recent steps are still turning
up something the run has not seen.

That replacement was a regression, and it shipped **demoted** rather than as
designed. The premise — "a step count means nothing on a run that can reach a
cube" — is false, because **every discovery run has tables**. A datasource that
cannot anchor may not run discovery on its own, so a cube is only ever present
*alongside* a warehouse. Replacing the floor therefore meant that connecting a
cube to an existing SQL project silently dropped its effective floor from sixty
steps to four, with no setting changed and nothing in the run log naming the
floor — because the floor never fired to be named.

So the rule now only ever **extends** a run past the floor
(`stopRule.acceptDone`, `services/agent/internal/ai/exploration_stopping.go`).
That is the safe ordering and it should stay. But it also means the original
complaint is untouched: the floor is still a scalar step count, and sixty
trivial slices still satisfy it.

## Shape of a fix

Make the floor count **steps that added something**, rather than steps.

The machinery already exists. `repeatsEarlierWork` scores each executed step
against the run's own vector index and `stopRule.observe` records the verdict —
today only to drive the extension rule. Feeding the same verdict into the floor
means a step that repeats earlier work does not advance the counter, and
"sixty steps" starts to mean sixty *distinct* ones.

## The cost, which is the actual decision

This makes spinning runs **strictly longer**, and it does so on the sources
where a step costs money. A model that re-asks the same question ten ways pays
for all ten and is credited with none, against a quota metered per request.
That is the correct incentive for the analysis and the wrong one for the bill,
[...]

PR #384, opened 2026-09-04
docs(agent): correct the stopping rule's header to match what it does

jale-dbox:
The doc comment at the top of `exploration_stopping.go` still describes the design as it was first drafted — the step floor *replaced* by a novelty rule on cube-reaching runs. That ordering was reversed before #382 merged, and the comment did not follow.

The reversal matters: every discovery run has tables, so a cube is only ever present alongside a warehouse, and replacing the floor meant connecting a cube to an existing project silently shortened the SQL half of its run. `acceptDone` gates on the floor first and lets novelty only extend past it. The header now says that.

It also names the part of the original problem that is still standing — the floor counts steps, not what they found — and points at #383, so the next reader does not have to reverse-engineer that from the code.

Comment only. No behaviour change, no test change.

## Verified

- `go build ./...` and `go vet ./internal/ai/` clean in `services/agent`
- `gofmt` clean

Part of #252.

— Co-coded with Jale 🤖

### DIFF
--- a/services/agent/internal/ai/exploration_stopping.go
+++ b/services/agent/internal/ai/exploration_stopping.go
@@ -15,14 +15,25 @@ package ai
 // termination the floor exists to prevent. The number is not the problem, and
 // no value of it is right.
 //
-// So on a run that can query a cube, the floor is replaced by what it was
-// standing in for: whether the exploration is still turning up anything new.
-// A step's novelty is its distance from the steps already taken, which the
-// run-scoped step index can answer because it is already embedding every step
-// for the analysis phase. While recent steps are still novel, a "done" signal
-// is rejected exactly as the floor rejected it; once a run of them says the
-// same thing as earlier work, the exploration has finished whether or not it
-// has taken a particular number of steps.
+// The floor is not replaced, though. Every discovery run has tables: a
+// datasource that cannot anchor may not run discovery on its own, so a cube is
+// only ever present ALONGSIDE a warehouse, and the floor remains a fair proxy
+// for those tables. Replacing it would mean that connecting a cube to an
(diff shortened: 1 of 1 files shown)


jale-dbox:
**Final report**

**What.** Rewrote the design paragraph at the top of `services/agent/internal/ai/exploration_stopping.go`. Comment only — no code, no tests, no behaviour.

**Why.** The header described the *first* version of #382: the step floor replaced by a novelty rule on cube-reaching runs. That ordering was reversed before merge — the floor gates first, novelty only extends past it — and the comment was left describing a design that no longer exists. It sat directly above an `acceptDone` whose own comment says the opposite, which is the worst place for a stale claim to live.

**How.** Three things the header now states: that the floor is *not* replaced and why (every discovery run has tables, so a cube is only ever alongside a warehouse); that the marginal-value rule is added past the floor and can only lengthen a run; and that the part of the original problem this did *not* solve — the floor counts steps, not what they found — is still open, pointing at #383.

**Verified.** `go build ./...`, `go vet ./internal/ai/`, `gofmt` clean in `services/agent`. CI green. One Codex round, clean: *"The diff only updates explanatory comments and does not change executable code or tests. I found no actionable correctness issues introduced by this patch."*

**Worth noting.** #382 ran thirteen review rounds and produced fifteen findings; none of them was this, and none of them was the regression that prompted the reversal either. Both were defects in what the code *said* rather than in what it *did*, and a reviewer checking code against its own comments will confirm agreement indefinitely without noticing that the pair are jointly wrong. I found this one by re-reading the merged file, not by reviewing it.

— Co-coded with Jale 🤖

PR #385, opened 2026-09-04
feat(models): ask for a table identifier only where tables exist (#252 W9)

jale-dbox:
The other half of W9's blocker, and a real prerequisite for W11 rather than the loose end I previously called it.

## The problem

`models.ValidateDomainPack` hard-required `{{DATASET}}` in every pack's exploration prompt (`domainpack_validator.go:84`). That placeholder is substituted with the datasource's dataset and schema names, so the prompt can say which tables exist to query. A cube has none — a query there is a choice of metrics and dimensions against a property, with no identifier to qualify them with.

So on a cube target the rule is not *unmet*, it is **unsatisfiable**. And it is not a distant concern: enterprise `packgen/validator.go:80` calls this function inside the synth auto-fix loop, so a generated cube pack re-rolls toward a placeholder its target has no use for until the retry budget is gone. W9's enterprise half removed the `{{DIALECT}}` demand; this is the same blocker on the other placeholder, and it bites the moment a cube prompt stops asking for SQL scaffolding — which is exactly what W11's exemplar has to teach.

## 1. The validator

`ValidateDomainPackForTarget(pack, target)` takes the shape from the caller. `ValidateDomainPack(pack)` keeps its signature and delegates with the pack's own `EffectiveShape()`.

The two entry points are not redundant, and the distinction is the thing W9's first enterprise attempt got wrong:

- **At save time** the pack's own shape is authoritative. The pack being persisted *is* the pack, and its shape is what selection keys on afterwards — a pack labelled cube is asked cube questions ever after, so the label and the relaxation travel together.
- **Mid-generation** it is not. The model's raw output never carries `shape` — it is not in the output schema and W8 strips it from the example pack — so reading the pack's own field there resolves every target to the default and leaves the rule exactly as unsatisfiable as before. It would also let a hallucinated `"shape": "cube"` switch the SQL checks off for a pack destined for a warehouse. The target is a fact; the pack's field is a claim.

Relaxed, not inverted. A cube pack carrying `{{DATASET}}` anyway still passes, and every other placeholder is required of a cube exactly as before. An unknown shape is treated as table-shaped, so a spelling this build has not heard of keeps the check rather than silently dropping it.

## 2. The hole that opens behind it (Codex R1, R2)

Making a cube pack savable makes it *selectable*. The domain picker lists every published pack, and generated packs are published — so the first cube pack in the corpus is offered to a customer creating a BigQuery project, whose prompts then tell the model to choose metrics and dimensions against a source that has tables. That fails without erroring, which is the expensive kind of failure.

So the pairing is now checked wherever it changes:

- **Project create**, where a pack and a datasource are first paired. The seeding block moves below the warehouse checks, because neither a split-brain body nor an unanchored project has a settled answer to which datasource the pack must agree with.
- **The settings edit** (`PUT /projects/{id}`), because a project may be created *before* it has a datasource — that is how the product is set up — and its pack is seeded then with nothing to disagree with. The datasource arrives one edit later, which is where the pairing first becomes checkable.

This is deliberately not a validator rule. A cube pack is not invalid for being cube-shaped; it is only wrong for a particular project.

The two guards fail in opposite directions, on purpose. `packShapeMismatch` reads an unregistered provider as table-shaped, so an unknown spelling keeps the check. The settings-edit wrapper lets a pack it cannot load through, because the pack may have been deleted or renamed since the project was created and blocking an unrelated settings edit on that is worse than the mismatch.

## Not in this PR

`packgen` must pass its resolved target for the validator change to take effect; on its own it is inert for the generator. That is the enterprise half and lands next. Stated plainly because the shape of the mistake here is a fix that looks complete and does nothing — which is precisely what W9's first attempt was.

The multi-datasource warehouse-management flow is not covered. It does not need to be today: a cube cannot anchor, a project's primary must be able to carry it, and the project-level pack is the primary's — so the primary stays table-shaped whatever is added beside it.

## Verified

- `go build ./...`, `go vet`, `gofmt` clean; `services/agent` and `libs/go-common` still build
- `go test ./...` across the **whole** `services/api` module, twice, not a `-run` filter
- 14 new tests. Validator: the cube exemption, the relaxation staying a relaxation, the SQL rule intact in both spellings of table-shaped, the caller's target beating a hallucinated one, an unknown target keeping the check, the exemption confined to `{{DATASET}}`. Pairing: both mismatched directions refused at create and at the settings edit, both matched directions accepted, an undeclared shape accepted (the entire existing corpus), an unresolvable provider keeping the check, a missing pack not blocking an edit
- Mutation-checked, eight mutants, all killed: predicate always-false, always-true (the bug as it stands today), the exported form reading a constant shape, the pairing guard never firing, an unknown provider skipping the check, reading `pack.Shape` instead of `EffectiveShape()`, the settings-edit guard removed, and a missing pack refusing the edit
- Three Codex rounds; two P2s, both valid, both fixed above; round 3 clean
- CI green

Part of #252.

— Co-coded with Jale 🤖

### DIFF
--- a/docs/concepts/domain-packs.md
+++ b/docs/concepts/domain-packs.md
@@ -46,6 +46,13 @@ that reads correctly and asks for queries the source cannot answer. A shape
 that is neither of the two above is rejected when the pack is saved, rather
 than stored and silently matching nothing.
 
+Shape also decides one validation rule. Every pack's exploration prompt must
+carry `{{SCHEMA_INFO}}` and `{{ANALYSIS_AREAS}}`, but `{{DATASET}}` — the
+dataset and schema names the prompt qualifies its tables with — is asked only
+of a pack written for an `entities` source. A cube has no tables to name, so
+requiring it there would be a rule no pack could satisfy. A cube pack that
+includes the placeholder anyway is still valid; nothing substitutes it.
+
 ## Three-Level Hierarchy
 
 ```
--- a/docs/concepts/domain-packs.md
+++ b/docs/concepts/domain-packs.md
@@ -53,6 +53,13 @@ of a pack written for an `entities` source. A cube has no tables to name, so
 requiring it there would be a rule no pack could satisfy. A cube pack that
 includes the placeholder anyway is still valid; nothing substitutes it.
 
+Shape also has to agree with the data source. Creating a project pairs a pack
+with a data source for the first time, and a pairing of different shapes is
+refused there: a cube pack seeded into a warehouse project produces prompts
+that ask for queries the source cannot answer, with no error to notice. A pack
+with no shape counts as `entities`, so every project created from the existing
+corpus is unaffected.
+
 ## Three-Level Hierarchy
 
 ```
--- a/services/api/internal/handler/project_pack_shape_test.go
+++ b/services/api/internal/handler/project_pack_shape_test.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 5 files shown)


jale-dbox:
**Final report**

**What.** `{{DATASET}}` is now asked only of a pack whose target names its tables, and the pack/data-source pairing is checked wherever it changes — at project create and at the settings edit that first attaches a data source.

**Why.** `packgen` runs this validator inside its synth auto-fix loop. A cube has no dataset or schema names, so the requirement was not unmet but **unsatisfiable**: every generated cube pack would re-roll toward a placeholder its target has no use for until the budget was gone. W9's enterprise half removed the `{{DIALECT}}` demand; this is the same blocker on the other placeholder, and it becomes live the moment a cube prompt stops asking for SQL scaffolding — which is what W11's exemplar exists to teach. I had previously called this a loose end that "doesn't bite today". Re-checking the call graph showed it is a W11 prerequisite.

**How.** `ValidateDomainPackForTarget(pack, target)` takes the shape from the caller; `ValidateDomainPack(pack)` delegates with the pack's own `EffectiveShape()`. Two entry points because the authoritative source differs: at save time the pack *is* the pack, and its label is what selection keys on afterwards; mid-generation the pack is the model's raw output, which never declares a shape — so reading its own field would leave the rule as unsatisfiable as before, and would let a hallucinated `"shape": "cube"` switch the SQL checks off for a warehouse pack.

**Verified.** `go test ./...` across the whole `services/api` module (not a `-run` filter); `services/agent` and `libs/go-common` still build; CI green. 14 new tests. Eight mutants, all killed — including the two that reproduce the bug as it stands today.

**Three Codex rounds, two P2s, both valid, both from the same blind spot.** I changed what a pack may *be* and did not ask what that makes *selectable*. R1: making a cube pack savable makes it appear in the domain picker, which lists every published pack, so it could be chosen for a BigQuery project. R2: the create-time guard I added for that cannot see the blank-project flow — a project created before it has a data source is seeded with nothing to disagree with, and the data source arrives one edit later. Round 3 clean.

Both are the shape of thing that gets found by asking who else reads this, and neither would have produced an error message if it had shipped — a cube pack on a warehouse project yields a discovery run that reads correctly and asks for queries the source cannot answer.

**Follow-up.** The enterprise half is decisionbox-io/decisionbox-enterprise#276, drafted: `packgen` must pass its resolved target or this change is inert for the generator, and it cannot compile until this merges, since `.platform-ref` names a branch. On its own, this PR is a validator that permits something nothing yet asks for.

— Co-coded with Jale 🤖

PR #386, opened 2026-09-04
feat(packs): add a cube-shaped reference pack (#252 W11)

jale-dbox:
The exemplar half of W11. The prompt half is decisionbox-io/decisionbox-enterprise#277; neither is much use without the other, and they can merge in either order.

## Why

Pack generation learns a pack's shape by imitating one published pack of the same shape (W8). Every pack in the corpus is table-shaped, so a cube source gets **no example at all** — the defined behaviour, and better than an example of the wrong shape, but it means the generator is writing a kind of pack it has never seen.

## What the pack is

Structurally cube, generic in domain, concrete in method — as the epic specifies.

**The method is the concrete part.** Every prompt teaches the same move: measure one metric, break it down by one dimension, find the value out of line with the rest of the breakdown, break that value down again. Six base analysis areas apply it to acquisition mix, engagement quality, conversion drop-off, audience composition, content performance and platform experience.

**The vocabulary is deliberately not any vendor's.** The pack says so in its own exploration prompt: *"Any metric or dimension named in this pack's own text is ordinary English, not a spelling from any particular source's API. Always take the exact name from the catalog."* The epic's reasoning for this is the part worth repeating — bleed between two analytics sources with similar-but-different names would look plausible and would not be caught, unlike the gaming-to-ecommerce bleed that was self-evident. The real vocabulary reaches the generator through W10's catalog slice; this pack supplies the reasoning.

It also states the two things about a cube that no table-shaped pack could teach: not every metric/dimension pairing is legal, so a refusal is an answer to route around; and breadth is not coverage, because the number of slices is the product of two long lists.

## Written for the window it is read through

A generator sees only the first few hundred characters of each prompt body — reference packs are trimmed so the example teaches structure, not content. So every prompt states the contract in its opening lines rather than in a section further down, and `TestCubeSeedPack_SaysItIsACubeInsideTheExcerptWindow` asserts that rather than trusting it.

That is also why this pack is a quarter the size of the hand-written ones. Front-loaded, not padded: text past the window is read by nothing.

## Published, and therefore not offered

Selection reads the published pool, so the pack must be published to be found. Publishing also means listed — and `GET /api/v1/domains` is the new-project domain picker. A project's pack describes its **primary** data source, which must be able to carry an analysis on its own, and no cube-shaped source can. Offering it would put an option in front of a customer that project creation then refuses.

So the picker now lists `entities` packs only. That also removes the per-data-source cube packs generation produces for enrichment sources, which were never domains to start a project from either.

## Verified

- `go test ./...` across the whole `services/api` module; `gofmt`, `go vet`, `go build` clean
- Six new tests: the pack passes the same contract the save and import handlers run; it declares its shape and is published; it carries no `{{DIALECT}}`/`{{DATASET}}`/`{{REF:}}` anywhere; every prompt names the cube contract inside the excerpt window; the picker drops a cube pack and keeps a table one; and a pack with no declared shape is still offered, which is the entire existing corpus
- Mutation-checked, four mutants, all killed: filter removed, filter inverted (which empties the picker for everyone — the reason that test asserts a table pack survives), the pack's `shape` field deleted, and a prompt padded so its contract falls outside the window

One property worth naming: deleting the `shape` field fails **two** tests, and the second is `IsSavable`. Without a declared shape the pack reads as `entities`, and an entities pack with no `{{DATASET}}` is rejected by the validator. The contamination risk W8 exists to prevent — a cube pack becoming the exemplar for SQL generation — is therefore structurally blocked for this pack rather than only tested for.

The epic's corpus-contamination criterion itself is already covered in the enterprise repo by `TestLoadReferencePacks_ACubePackInTheCorpusLeavesATableExampleUnchanged`, which asserts byte-for-byte against a real Mongo. Not duplicated here.

## Corrected in review

Four defects in the pack, none of which any test I had written would have caught, because they were all wrong statements rather than wrong code:

- **It taught reasoning and never the reply format.** Every table-shaped pack teaches the one-JSON-object-per-turn action envelope; this one taught none of it, so a pack generated by imitating it inherited the omission.
- **Its recommendation example was wrong where it overlapped the model** — `"priority": "high"` against `Priority int`, `expected_impact` as a sentence against an `Impact` object. Both decode to zero values with no error raised anywhere. Now derived from the struct by a test, and decoded into it by another.
- **It said `{{SCHEMA_INFO}}` holds this source's catalog, and that `search_tables` was not used here.** Both false, and the second follows from the first. `{{SCHEMA_INFO}}` is rendered from **table** schemas, so a cube contributes nothing to it; `search_tables` is the only runtime surface carrying its metric and dimension names. A pack teaching otherwise leaves the model to invent a name — a request that spends quota and returns nothing.
- **It carried `{{PROFILE}}` in the exploration prompt**, which only the base-context half substitutes. It reached the model as literal braces under an empty heading. No other seed pack does this; the pack had copied packgen's generation rule rather than its siblings. That rule is a pre-existing defect affecting SQL packs too — filed as decisionbox-io/decisionbox-enterprise#282 rather than changed here, since changing it changes what every customer's pack is generated from.

Each is now pinned by a test, including one that walks every `{{TOKEN}}` in the exploration prompt against the set the orchestrator actually replaces there.

Part of decisionbox-io/decisionbox-platform#252.

— Co-coded with Jale 🤖


jale-dbox:
**Codex round 1 — two P2s, both real defects. Fixed. But the reasoning for the first one does not hold, and the difference matters.**

> **[P2] Add the exploration action contract** — `seed/digital-analytics.json:31`
> When a cube pack seeded from this reference is used for discovery, the exploration loop has no structured output and `ParseAction` only accepts JSON actions with keys like `query`, `search_tables`, or `done`. This prompt explains how to choose metrics/dimensions but never tells the model to emit that envelope or to put the cube report request in the `query` string; the fallback repair nudge is SQL-shaped, so such runs can fail parsing or drift into `SELECT` queries. Add a cube-specific "How to explore" action-format section.

> **[P2] Use the full recommendation output schema** — `seed/digital-analytics.json:32`
> For LLMs/providers that follow the prompt rather than the advisory response schema, this example asks for only `title`, `description`, string `priority`, `related_insight_ids`, and a string `expected_impact`. The recommendation model and dashboard expect structured impact fields plus `category`, `target_segment`, `segment_size`, `actions`, and `confidence`, so recommendations generated from this pack will parse with empty metadata/action steps and no impact metric/improvement. Mirror the existing seed recommendation schema here.

## The correction

Both findings open with "when this pack is used for discovery". A cube pack is never used for discovery through these prompts, and it cannot be. A cube source cannot anchor, so it is never a project's primary; a non-primary datasource contributes only `datasourceFocusAreas(d.prompts)` — its enabled **analysis-area names** — to the run (`orchestrator_multiwarehouse.go:250`). Its exploration, base_context and recommendations prompts are never rendered. The primary's pack supplies those.

So neither defect can produce the failure described. I checked before fixing, because "it would break discovery" and "it teaches a generator to write a pack that would break discovery" call for different fixes, and only the second one is true here.

## Why they are still defects, and fixed

**The action contract.** Every table-shaped pack teaches the one-JSON-object-per-turn envelope; this one taught none of it. The exemplar is imitated, so the omission propagates into every generated cube pack — and the moment a per-datasource pack's prompts *are* rendered, which the fields exist for, it becomes live. Added a `How To Explore` section: `query` with `datasource_id` (a cube is always connected alongside a warehouse, so an unrouted query goes to the wrong place), `done`, and why `lookup_schema`/`search_tables` do not apply here.

**The recommendation schema.** This one is wrong today, not merely thin. `"priority": "high"` is a string against `Priority int`; `expected_impact` was a sentence against an `Impact` object. Both decode to zero values with **no error raised anywhere** — the dashboard renders priority 0 and an empty impact, and nothing in the pipeline says why. Now mirrors `models.Recommendation` exactly.

## What I did not take

The suggestion to "mirror the existing seed recommendation schema" — those packs also teach `effort` and `success_metrics`, which the model does not read. The example is now derived from the struct instead, by a test.

The `query` payload stays *described* rather than shown. Its format belongs to whichever source is connected, and writing one vendor's request shape into a pack the epic requires to be domain-generic is precisely the bleed it warns about: "bleed between adjacent cube sources — analytics properties with similar-but-different vocabularies — would be plausible-looking and would not [be caught]".

## Guarded, not just fixed

- `TestCubeSeedPack_RecommendationExampleMatchesTheModel` derives the required field names from `models.Recommendation` and `models.Impact` by reflection, so **adding** a field to the model fails until the pack follows
- `TestCubeSeedPack_RecommendationExampleParsesAsTheModel` decodes the fenced example into that struct — the type half, which a field-name check cannot see
- `TestCubeSeedPack_TeachesTheActionEnvelope` pins the envelope keys, and that no SQL appears inside it: the envelope is shared with the warehouse packs, its payload is not

Three more mutants, all killed: priority back to a string, `expected_impact` flattened to a sentence, and the `How To Explore` section deleted. Seven mutants killed on this PR now.

## One thing outside this PR

`buildDatasourcesPromptSection` opens with *"This project has multiple SQL datasources"* and illustrates routing with two SQL statements (`orchestrator_multiwarehouse.go:226-232`). That is rendered on every multi-datasource run — which is every run a cube can appear in — so the model is told the cube datasource is SQL. Unlike the two above, that one is live. Filing it separately rather than growing this PR into the agent's routing contract.

— Co-coded with Jale 🤖

jale-dbox:
**Round 4 — one P2, and it is a trap my own two guards built between them. Fixed.**

> [P2] Enforce cube-pack exclusion at project creation — `domains.go:53-55`
> Filtering here only protects the picker. A client can still `POST /api/v1/projects` with `domain: "digital-analytics"` before configuring a warehouse; `ProjectsHandler.Create` loads the seeded cube pack by slug and `packShapeMismatch` returns no mismatch when the primary provider is empty, so the project is persisted with cube prompts and later adding any anchoring entities datasource is rejected.

Correct, and the end state is worse than "unusable": it is **unrepairable through the API**. The project holds a cube pack, and the settings-edit guard then refuses every anchoring data source it could ever be given — the only kind it is allowed to have.

Neither guard is wrong on its own. `packShapeMismatch` returned "" for an empty data source on the reasoning I wrote into its comment: *"nothing has been chosen to disagree with."* The settings-edit guard refuses a mismatched pairing when the data source arrives. Put together, the first creates the state the second forbids.

The reasoning was the mistake. The pack has to agree with whatever data source the project ends up with; that source must be able to carry the analysis on its own, and no cube can — so the answer is knowable before the source exists. It is also the rule the picker already applies, which should have been the clue that the create path could apply it too.

Now: an absent data source is checked against `entities`, the refusal no longer describes a data source the project does not have, and the ordinary case — creating a project before connecting anything, which is how the product is set up — is asserted to still pass, since a guard that overreached here would break every project create in the product.

Mutation-checked: restoring the `primary.Provider == ""` early return fails both new tests. Eight mutants killed on this PR.

Round 5 running.

— Co-coded with Jale 🤖

jale-dbox:
**Round 5 — one P2, and it is the same hole one level up. Fixed.**

> [P2] Run cube-shape checks outside prompt seeding — `projects.go:799-803`
> When a create request supplies any custom `prompts`, `Create` skips the domain-pack lookup because the only call to `packShapeMismatch` is inside `p.Prompts == nil`. That means a client can still POST the newly seeded cube slug with no warehouse and custom prompts, creating the same cube-domain/no-datasource project this branch is trying to reject.

Correct. I fixed the *reasoning* about the empty data source last round and left the guard where it was — inside the seeding branch — so the whole refusal remained skippable by sending a `prompts` field. Supplying custom prompts does not make a project's domain compatible with its data source; it only stops the pack being copied into it. The pack lookup now happens whenever there is a `domainPackRepo`, and the pairing is checked on it either way.

**What I deliberately did not change alongside it:** an unknown domain is still refused only when the request needed the pack to seed from. This route has never looked a domain up for a client bringing its own prompts, and making it start now would be a second change riding along with the fix — one that would break any caller using a domain string the corpus does not have. `TestCreate_CustomPromptsWithAnUnknownDomainStillPass` asserts it did not happen, and the mutant that makes it happen fails that test.

Two more mutants, both killed: the check moved back inside the seeding branch, and the unknown-domain refusal widened. Ten on this PR.

Round 6 running.

— Co-coded with Jale 🤖

jale-dbox:
**Round 7: clean.** *"The changes consistently filter cube-shaped packs from the project domain picker and enforce pack/source shape compatibility during project creation without introducing an evident correctness regression."* CI green, `MERGEABLE/CLEAN`.

(Round 6 was also clean, but a lint fix landed after it — staticcheck `QF1002` on a tagless switch, which `go vet` does not check and I had not run `golangci-lint` before pushing. A clean round only covers the diff it saw, so it was re-run rather than assumed.)

**Final report.**

**What.** A cube-shaped reference pack, `digital-analytics`, plus the two guards that publishing it requires.

**Why.** Pack generation learns a pack's shape by imitating one published pack of the same shape (W8). Every pack in the corpus was table-shaped, so a cube source got no example at all — the defined behaviour, better than an example of the wrong shape, but it left the generator writing a kind of pack it had never seen.

**The pack.** Structurally cube, generic in domain, concrete in method, as the epic specifies. Every prompt teaches one move: measure a metric, break it down by a dimension, find the value out of line with the rest, break that value down again. Six base areas apply it. The vocabulary is ordinary English rather than any vendor's API spelling, and the pack says so — the bleed the epic warns about is between adjacent cube sources with similar-but-different names, which would look plausible and would not be caught. It states the two things no table-shaped pack could teach: not every metric/dimension pairing is legal, and breadth is not coverage.

Written for the window it is read through. A generator sees only the first few hundred characters of each prompt, so each states its contract in its opening lines, and a test asserts that rather than trusting it. That is why the pack is a quarter the size of the hand-written ones — front-loaded, not padded.

**The guards.** Publishing is what makes a pack findable by selection, and also what puts it in the new-project domain picker. A project's pack describes its **primary** data source, which must carry the analysis on its own, and no cube can — so the picker lists `entities` packs only, and project creation refuses a mismatched pairing whether or not the request brought its own prompts.

**Seven rounds. Six findings, every one a false statement rather than broken code:**

1. The pack taught reasoning and never the reply format — no action envelope, which every table-shaped pack has.
2. Its recommendation example was wrong where it overlapped the model: `"priority": "high"` against `Priority int`, `expected_impact` as a sentence against an object. Both decode to zero values with nothing raised.
3. It claimed `{{SCHEMA_INFO}}` holds this source's catalog and that `search_tables` was unused. Both false: `{{SCHEMA_INFO}}` is rendered from **table** schemas, and `search_tables` is the only runtime surface carrying a cube's names. I had conflated packgen's generation-time slice with the generated pack's run-time placeholder.
4. It carried `{{PROFILE}}` in exploration, which only the base-context half substitutes — literal braces under an empty heading.
5. A cube pack could be created with **no** data source, because my own guard read the empty case as "nothing to disagree with". The project then held cube prompts and the settings-edit guard refused every anchoring source it could ever be given: unusable and unrepairable, built out of two guards disagreeing rather than either being wrong.
6. That fix left the check inside the seeding branch, so sending a `prompts` field skipped it entirely.

Findings 5 and 6 are the ones worth remembering. Both were holes I created, both in the *empty* and *skipped* cases rather than in the behaviour I was thinking about, and neither would have raised an error — one produces a project nobody can repair, the other quietly does nothing.

**Verified.** Full `services/api` suite; `golangci-lint run ./...` clean; ten mutants killed, including the two that restore findings 5 and 6. Eleven tests, several of which derive their expectations from the model or the template rather than restating them, so drift fails rather than passes.

**Filed, not fixed:** decisionbox-io/decisionbox-platform#387 (routing contract calls every datasource SQL — live, unlike the two pack findings review raised in round 1) and decisionbox-io/decisionbox-enterprise#282 (`{{PROFILE}}` required of every generated pack and substituted in none).

**Pairs with** decisionbox-io/decisionbox-enterprise#277, which stops the prompt asking a cube target for SQL. Either order.

— Co-coded with Jale 🤖

Issue #387, opened 2026-09-04
Multi-datasource routing contract tells the model every datasource is SQL


jale-dbox:
## The problem

`buildDatasourcesPromptSection` is rendered on every multi-datasource discovery run, and it opens by asserting something that is not true of all of them:

```go
b.WriteString("This project has multiple SQL datasources. Each SQL statement runs against exactly ONE datasource: ...")
```

`services/agent/internal/discovery/orchestrator_multiwarehouse.go:226`

The two worked examples that follow (lines 231–232) are both SQL statements, one of them a `SELECT ... JOIN` shown as the wrong way and a two-step `WHERE id IN (...)` hop shown as the right way.

A cube-shaped datasource has no SQL and no tables. It is also **only ever** present on a multi-datasource run — a source that cannot anchor may not run discovery on its own — so this section is exactly the text a model sees when one is connected. It is told, in the same breath, that the property it is about to query is a SQL datasource, and shown two SQL statements as the model to follow.

## Why the rest of the pipeline does not save it

The pieces that know better sit on either side of this text and do not reach it:

- The per-datasource descriptor loop below (line 238) prints provider, table count, subject areas and key metrics — but never the datasource's **query language**, which `QueryRunner.QueryLanguage()` already exposes and `queryexec` already resolves per datasource.
- `NonSQLLanguage` reaches the *repair* path (`ai/sql_fixer.go:82`), so a query that fails is corrected in the right language. Nothing corrects the framing that produced it.
- A cube datasource's own domain pack contributes only its analysis-area **names** here (`datasourceFocusAreas`, line 250), so nothing it says about its own shape reaches the run either.

## Impact

A model told a source is SQL will write SQL for it. That is a query the source rejects, one wasted step against a source metered per request, and a repair loop entered for a reason the run created itself. The failure is loud rather than silent, which is the better direction — but it is a per-step tax on exactly the sources where steps cost money.

## Shape of a fix

The routing contract needs to be written per datasource rather than asserted once for all of them:

- Say "multiple datasources", not "multiple SQL datasources".
- Print each datasource's query language in its descriptor block, from the language `queryexec` already resolves.
- Give the hop example a non-SQL side, or state the hop rule in terms of values passed between steps rather than in terms of `WHERE ... IN`.

The last point is the one with a design question in it: the hop rule is currently taught by example, and a two-language example is longer than a one-language one on a prompt that is already long.

## Not urgent

[...]

PR #390, opened 2026-09-07
fix(agent): stop telling the model every datasource is SQL

jale-dbox:
## What

The multi-datasource routing contract stops telling the model that every datasource on the run is SQL.

`buildDatasourcesPromptSection` opened with *"This project has multiple SQL datasources. Each SQL statement runs against exactly ONE datasource…"* and then taught the hop rule with two SQL statements — a cross-datasource `JOIN` as the wrong way, a `WHERE id IN (…)` two-step as the right way. A source with no SQL is **only ever** present on a multi-datasource run, since one that cannot anchor may not run discovery on its own, so this is precisely the text a model reads when such a source is connected. Told the thing it is about to query is SQL, it writes SQL: a rejected query, a step spent against a source metered per request, and a repair loop entered for a reason the run created itself.

## How

The contract is split on the **query language**, resolved from the registry by provider slug:

- **All-SQL run** — renders byte for byte as it did. That is every multi-warehouse project that exists today, and the sentence is true for them.
- **A run carrying a non-SQL datasource** — gets a contract that says "multiple datasources", names each datasource's language, scopes `lookup_schema` to the datasources that have tables, states the hop's second step as *a filter written in B's own query language* rather than as a `WHERE … IN` clause, and claims precedence over the query examples above it.

Per datasource, on a mixed run:

```
- **`default`** (Sales PG) — postgres, 4 tables
  - Query language: PostgreSQL
- **`wh_analytics`** (Web Analytics) — ga4, NO TABLES: a metric/dimension cube. `lookup_schema` does not apply to it; `search_tables` lists its metrics and dimensions.
  - Query language: GA4 Report Request (JSON) — not SQL, which this datasource rejects
```

Every datasource is named, SQL ones included — *"one of these is different"* is not something a model can act on. And a source with no tables says so rather than reporting `0 tables`, a count that reads as an empty or broken warehouse and invites the one action that cannot work against it.

Four decisions worth the reviewer's attention:

- **Split on the language, not the shape.** A record system with tables and its own query API is entity-shaped; shape says nothing unusual about it, and calling it SQL is just as wrong as it is for a cube. `NonSQLLanguageOf` is also the answer that survives middleware — a wrapper that does not re-expose the query seam erases a live provider's declaration, and a registration cannot be wrapped.
- **`ProviderMeta.Language()` is not consulted first.** It falls back to `"SQL"` for a source that declares a shape and no language — exactly the source this section exists to stop mislabelling.
- **The worked example stays SQL-to-SQL, and says so.** Writing a second one in some specific non-SQL request format would teach a syntax this package has no provider for and no way to keep true. The rule above it is stated in terms of the values passed between steps, so it holds either way.
- **The section claims precedence over what precedes it.** What precedes it is the *project's* exploration prompt — its primary datasource's domain pack — which teaches one action contract with worked examples in one language, and a non-primary datasource's own pack contributes only its analysis-area names to a run. So on a mixed run the model reads a full action contract in one language and then a list of datasources that do not all speak it. This section previously reinforced that framing; it now says which one wins and what actually differs (the action and its `datasource_id` are unchanged; only the `query` field's content does). Stated as precedence rather than as "SQL loses", so it stays true whichever language the project's pack was written in.

## Verification

- `go test ./...` in `services/agent` — green; `golangci-lint run ./internal/discovery/...` — 0 issues. CI green.
- **Byte-identity for every existing project is proven by the diff, not asserted.** The first commit captures the section as it rendered *before* any edit into `testdata/datasources_prompt_all_sql.golden`; no later commit touches that file. The golden is a fixture, not an output — there is no `-update` flag, because regenerating it is the one way to lose the evidence.
- **12 mutants, 12 killed.** A mutation pass over the new code left two survivors, both fixed by new tests in `257d4bc`: dropping the shape check from the no-tables line (a table-shaped datasource whose cached schemas came back empty would be described as a metric/dimension cube), and deleting the sentence that scopes `lookup_schema` (the cube's own line said it, so every test still passed while the contract paragraph no longer did). Both were gaps in the tests, not in the code.
- Tests cover: the switch itself (one non-SQL datasource rewrites the contract, nothing else does); every language named on a mixed run; a cube not reporting `0 tables`; a cube that *does* arrive carrying tables keeping the ordinary line, because the model can see them; a table-shaped source with zero tables not being called a cube; the hop rule surviving two languages; the precedence claim; and the language resolver, including a cube that declares its language nowhere — the case `ProviderMeta.Language()` answers `"SQL"` for.

## Not done here

**#391** — a mixed run still renders the primary's action contract, worked SQL examples and all, to every datasource. The precedence sentence above is the most a section appended at the end can honestly do; actually gating or rewriting that contract per datasource changes how a run assembles its prompt, has three viable designs and no obviously right one, and is filed rather than guessed at. Raised as a P2 in Codex round 2.

`ask_serve.go:254` resolves `capability.QueryLanguage = meta.Language()`, which has the same fallback this PR avoids: a cube declaring neither a query language nor a display dialect would reach the ask prompt as `"SQL"`, and `writeCubeCatalogLine` would render *"Query language: SQL (not SQL)"*. Not reachable with any provider that exists — GA4 declares both — and a different prompt on a different path.

The file header's `docs/design/multi-warehouse.md` reference points at a document that is not in this repo. Pre-existing, left alone. No public doc describes this prompt, so nothing under `docs/` went stale.

Closes #387

— Co-coded with Jale 🤖

### DIFF
--- a/services/agent/internal/discovery/datasources_prompt_language_test.go
+++ b/services/agent/internal/discovery/datasources_prompt_language_test.go
@@ -1,0 +1,50 @@
+package discovery
+
+import (
+	"os"
+	"testing"
+
+	"github.com/decisionbox-io/decisionbox/services/agent/internal/models"
+)
+
+// sqlOnlyContext is the all-SQL multi-datasource run: the shape every
+// multi-warehouse project has today. It exercises every field the routing
+// contract renders — label, description, card, per-datasource pack areas — so
+// the golden below covers the whole section rather than its opening.
+func sqlOnlyContext() *datasourceContext {
+	return &datasourceContext{descriptors: []datasourceDescriptor{
+		{
+			id: "default", label: "Sales PG", provider: "postgres", tableCount: 4,
+			description: "orders and invoices",
+			card: &models.WarehouseCard{
+				SubjectAreas: []string{"sales", "billing"},
+				KeyEntities:  []string{"invoice", "customer"},
+				KeyMetrics:   []string{"revenue", "arpu"},
+			},
+		},
+		{
+			id: "wh_oracle", label: "Catalog", provider: "oracle", tableCount: 7,
+			description: "music catalog",
+			prompts: &models.ProjectPrompts{AnalysisAreas: map[string]models.AnalysisAreaConfig{
+				"g": {Name: "Genre Performance", Enabled: true},
+			}},
+		},
+	}}
+}
+
+// TestBuildDatasourcesPromptSection_AllSQLIsUnchanged pins the text every
+// existing multi-warehouse project receives, byte for byte.
+//
+// The golden was captured from the tree before this file existed, so a diff
+// against it is a diff against what shipped. It is a fixture, not an output:
+// regenerating it to make a failure go away deletes the only evidence that a
+// run with no non-SQL datasource still reads exactly as it did.
+func TestBuildDatasourcesPromptSection_AllSQLIsUnchanged(t *testing.T) {
+	want, err := os.ReadFile("testdata/datasources_prompt_all_sql.golden")
+	if err != nil {
+		t.Fatalf("read golden: %v", err)
+	}
+	if got := buildDatasourcesPromptSection(sqlOnlyContext()); got != string(want) {
+		t.Errorf("all-SQL routing contract changed.\n--- got ---\n%s\n--- want ---\n%s", got, want)
+	}
+}
--- a/services/agent/internal/discovery/datasources_prompt_language_test.go
+++ b/services/agent/internal/discovery/datasources_prompt_language_test.go
@@ -1,11 +1,53 @@
 package discovery
 
 import (
+	"errors"
 	"os"
+	"strings"
 	"testing"
 
+	gowarehouse "github.com/decisionbox-io/decisionbox/libs/go-common/warehouse"
 	"github.com/decisionbox-io/decisionbox/services/agent/internal/models"
 )
+
+// Providers registered only for these tests, so language resolution runs
+// against the real registry — the thing the running process reads.
+//
+// Registered from init() rather than from a test body: the registry is
+// process-global and panics on a duplicate slug, so a registration inside a
+// test passes once and takes the package down under `go test -count=2`.
+const (
+	testSQLSlug        = "discovery-lang-sql"
+	testCubeNamedSlug  = "discovery-lang-cube-named"
+	testNativeRowsSlug = "discovery-lang-native-rows"
+)
+
+func init() {
+	factory := func(gowarehouse.ProviderConfig) (gowarehouse.Provider, error) {
+		return nil, errors.New("registered for its metadata only")
+	}
+	// An ordinary SQL warehouse: a display dialect and no language of its own.
+	gowarehouse.RegisterWithMeta(testSQLSlug, factory, gowarehouse.ProviderMeta{
+		Name: "Lang SQL", Dialect: "PostgreSQL",
+	})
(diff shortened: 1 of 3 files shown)


jale-dbox:
## Final report

**What.** The multi-datasource routing contract no longer asserts that every datasource on the run is SQL. It now splits on the query language: an all-SQL run renders byte for byte as before, and a run carrying a datasource that is not SQL gets a contract that names each datasource's language, scopes `lookup_schema` to the datasources that have tables, states the hop's second step as a filter in the target's own language, and claims precedence over the query examples above it.

**Why.** A source with no SQL is only ever present on a multi-datasource run — one that cannot anchor may not run discovery alone — so this section is exactly what a model reads when one is connected. Being told the thing it is about to query is SQL, it writes SQL: a rejected query, a step spent against a source metered per request, and a repair loop entered for a reason the run created itself. Loud rather than silent, which is the better direction, but a per-step tax on the sources where steps cost money.

**How.**

- `runHasNonSQLDatasource` reads `NonSQLLanguageOf` per provider slug. Language, not shape — a record system with tables and its own query API is entity-shaped, and calling it SQL is just as wrong. Registry, not live provider — a wrapper that does not re-expose the query seam erases a live declaration; a registration cannot be wrapped.
- `datasourceQueryLanguage` asks `NonSQLLanguageOf` *before* `ProviderMeta.Language()`, because the latter falls back to `"SQL"` for a source that declares a shape and no language — precisely the source this section exists to stop mislabelling.
- The all-SQL text moved verbatim into `writeSQLRouting`; the hop example is shared by both contracts so they cannot drift on the part neither disputes.
- A datasource with no tables says so instead of reporting `0 tables`, and the claim is read off the run as well as the registry — a cube that somehow arrives carrying tables keeps the ordinary line, because the model can see them.

**How it was verified.**

- `go test ./...` in `services/agent` green; `golangci-lint run ./internal/discovery/...` 0 issues; CI green on `f88868a` (`MERGEABLE/CLEAN`).
- Byte-identity for existing projects is proven by the diff, not asserted: the first commit captures the section as it rendered *before* any edit, and no later commit touches that golden.
- **Mutation pass: 12 mutants, 12 killed.** Two survived the first round and are now covered — a table-shaped datasource whose cached schemas came back empty being described as a metric/dimension cube, and the mixed contract silently losing the sentence that scopes `lookup_schema` (the cube's own line said it, so nothing failed). Both were test gaps, not code bugs.
- **Codex: 3 rounds.** R1 clean. R2 raised one P2 — the section is appended after the project's own SQL-shaped exploration prompt, so the final prompt still contradicts itself. R3 clean after the precedence fix.
- Copilot did not review this PR; there are no review threads to resolve.

**Follow-ups.**

- **#391** — the P2's other half. A mixed run still renders the primary's action contract, worked SQL examples and all, to every datasource. The precedence sentence is the most a section appended at the end can honestly do; gating or rewriting that contract per datasource changes how a run assembles its prompt, has three viable designs and no obviously right one. Filed rather than guessed at.
- `ask_serve.go:254` has the same `meta.Language()` fallback this PR avoids: a cube declaring neither a query language nor a display dialect would reach the ask prompt as `"SQL"`, rendering *"Query language: SQL (not SQL)"*. Not reachable with any provider that exists, and a different prompt on a different path — noted, not fixed.

**Known limitation.** Nothing here is exercised by a real run: the only non-SQL provider is gated off by default, so every assertion about the mixed path is a test against a registered stub, not against GA4. That is the same limitation the rest of Phase 1 carries.

— Co-coded with Jale 🤖

PR #292, opened 2026-06-24
ci(docker): push decisionbox-agent image to ECR alongside GHCR

jale-dbox:
## Summary

- Add ECR push steps to the Docker publish workflow for the `decisionbox-agent` image
- After the existing GHCR multi-arch push, the workflow authenticates to AWS via OIDC, logs into ECR, and copies the multi-arch manifest using `docker buildx imagetools create`
- Steps are gated on `matrix.service == 'decisionbox-agent'` — the API and dashboard images are unaffected

## Changes

| File | What changed |
|------|-------------|
| `.github/workflows/docker-publish.yml` | Added `id-token: write` permission + 3 new steps (AWS creds, ECR login, ECR push) gated on agent matrix entry |
| `CHANGELOG.md` | Added entry under `[Unreleased]` |
| `docs/deployment/kubernetes.md` | Note ECR image URI as alternative for AWS/EKS |
| `docs/reference/configuration.md` | Mention ECR mirror in both `AGENT_IMAGE` entries |

## Design decisions

- **`docker buildx imagetools create`** instead of `docker tag` + `docker push`: copies the multi-arch manifest (amd64+arm64) cross-registry without re-pulling layers locally. The workflow already pushes multi-arch to GHCR, so this preserves the manifest list as-is.
- **Official `aws-actions/amazon-ecr-login@v2`** instead of raw `aws ecr get-login-password`: handles token refresh and error reporting cleanly.
- **Agent-only conditional**: the issue scopes this to `decisionbox-agent` only. Extending to other images later is a one-line conditional change.
- **ECR steps run after GHCR push**: if ECR fails, GHCR images are already published (acceptable partial state).

## Required secrets (pre-configured)

| Secret | Purpose |
|--------|---------|
| `AWS_DEPLOY_ROLE_ARN` | OIDC role assumption |
| `AWS_ACCOUNT_ID` | ECR registry prefix |

## Test plan

- [x] YAML validates (parsed locally)
- [ ] Push to `main` or tag → workflow shows "Configure AWS credentials" / "Login to Amazon ECR" / "Push to ECR" steps as green for agent, skipped for api/dashboard
- [ ] `docker manifest inspect <account>.dkr.ecr.us-east-1.amazonaws.com/decisionbox-agent:<tag>` shows amd64+arm64

Closes #291

— Co-coded with Jale 🤖

PR #395, opened 2026-09-08
feat(warehouse): three-legged auth, a provider-keyed app registration, and a credential a datasource can name

jale-dbox:
**Part of #252** (enterprise decisionbox-io/decisionbox-enterprise#252), platform half of enterprise **#292**. Targets `feat/non-sql-datasources`.

The community half of user-delegated OAuth for datasources: what the registry has to be able to say, and what the agent has to be able to read, so the enterprise side can hold a grant on a connection.

## What this adds

**A three-legged auth method.** `AuthMethod` could only describe a form to type into. It gains a `Flow`, and an `AuthorizationCode` block carrying where to send the user, where to redeem the code, where to end the grant, and the scopes the consent must cover — the last because providers with granular consent let a user approve some and withhold others, so it is what the exchange checks the grant *against*, not merely what it asks for.

**The OAuth app registration, keyed by provider.** The deployment's own client was stored under a key derived from the datasource slug, which quietly asserts that every consumer needs its own registration. It does not: a customer who registers one Google client should use it for every feature that authenticates against Google, and never register a second. So the key becomes the OAuth provider, which each consumer declares in its registry entry — both registries gain that declaration, and since neither owns the key scheme it moves to its own package (`libs/go-common/oauthreg`).

The keys are readable now (`oauth-app-google-client-id`) rather than hex. The hex existed to keep arbitrary slugs inside Azure Key Vault's alphabet; with both halves drawn from a closed set of compile-time constants, a lossy substitution is a collision a reviewer can see rather than one a customer can cause.

**A credential a datasource can name.** `initWarehouseProvider` derived the secret key from the datasource id, which cannot reach a credential that several datasources share. A datasource may now name its slot (`credential_ref`); when it does not — every datasource that exists today — nothing about the read changes.

**Loud failures where silence would mislead.** A three-legged method that names no OAuth provider can never authenticate, so it is a registry declaration error rather than a generic "not configured" from the provider. An unusable `credential_ref` is refused rather than falling back to the derived key, which would look like a working connection authenticating as something else.

**Where a grant's account is reported.** A three-legged method could say what a grant must *cover* but not who authorized it, so a connection made through one had nothing to be labelled with but its provider's name — two connections to the same source were indistinguishable. `AuthorizationCode` gains `UserInfoURL` and `IdentityScopes`, declared alongside `RevokeURL` because they are facts the provider publishes rather than deployment configuration.

`IdentityScopes` stay out of `Scopes` deliberately: the exchange requires everything in that list, and a withheld display name must not cost a working grant.

**An extension point on the connect form.** A three-legged method's notice says to authorize the data source *after* saving it, which is the honest answer only where nothing better is on offer. `WarehouseFormFields` gains an optional `authorizationSlot` that replaces the notice when a caller supplies one, so a build that can collect the authorization on the form itself is not forced to render its picker beneath a notice telling the customer to go elsewhere. Absent the prop, nothing changes.

## What is deliberately not here

- Anything Google-specific. The whole surface is declared by whichever provider adopts it.
- A consent URL builder. The code is obtained in the browser by Google Identity Services; a redirect leg ships with the provider that needs one (Salesforce).
- Credential storage or revocation. Both are enterprise-side, on the connection that holds the grant.

## Verification

- `libs/go-common` + `services/agent` + `services/api` unit suites, and the composed dashboard suite (737 tests).
- The credential-key change is pinned by a characterization test written **before** it, so "the derived path still works" is a test that was already passing rather than one written to agree with the new code.
- Mutation-checked: ignoring the ref, leaving it in the provider config, accepting an unusable one, and treating an empty one as set are all caught.
- `golangci-lint` clean on every touched package.

## Known limitation

Nothing reads `oauth-app-*` until the enterprise side writes it, so this merges as inert plumbing — which is the point: enterprise CI cannot compile its half until this lands on the bridge branch.

— Co-coded with Jale 🤖

### DIFF
--- a/libs/go-common/connectors/registry.go
+++ b/libs/go-common/connectors/registry.go
@@ -18,9 +18,21 @@ type Factory func(cfg Config) (Connector, error)
 // ProviderMeta describes a connector for UI rendering (the instance-admin
 // app-registration form).
 type ProviderMeta struct {
-	ID           string        `json:"id"`
-	Name         string        `json:"name"`
-	Description  string        `json:"description"`
+	ID          string `json:"id"`
+	Name        string `json:"name"`
+	Description string `json:"description"`
+
+	// OAuthProvider identifies whose app registration this connector
+	// authenticates with — "google", say. Declared rather than assumed from the
+	// connector kind, because the registration is shared: a deployment that has
+	// registered a Google client for one feature must not be asked to register
+	// a second one here. Empty for a connector that needs no OAuth app.
+	OAuthProvider string `json:"oauth_provider,omitempty"`
+
+	// ConfigFields are the connector's OWN instance-level fields — what it
+	// needs beyond the shared OAuth registration. The Google Picker's api_key
+	// and app_id are the example: Drive needs them, and a consumer of the same
+	// Google client that never opens a picker does not.
 	ConfigFields []ConfigField `json:"config_fields"`
 }
 
--- a/libs/go-common/oauthreg/oauthreg.go
+++ b/libs/go-common/oauthreg/oauthreg.go
@@ -1,0 +1,79 @@
+// Package oauthreg names the deployment's own OAuth app registration: the
+// client_id, client_secret and redirect_uri an operator registers once with a
+// provider, so that no DecisionBox-owned application ever holds a customer's
+// data grant.
+//
+// # Why this is not part of either registry
+//
+// Two unrelated registries have consumers that authenticate against the same
+// provider — a data source in the warehouse registry, an external-storage
+// connector in the connectors one. Both need the same three fields, and a
+// customer who has registered a Google client should not have to register a
+// second one because the feature that needs it lives in a different registry.
+//
+// So the registration is keyed by the OAuth provider ("google"), not by the
+// consumer, and the key scheme lives in a package neither registry owns. Which
+// provider a consumer belongs to is declared in its own registry metadata; this
+// package only knows how the resulting key is spelled.
+package oauthreg
+
+import "strings"
+
+// Registration fields. A three-legged flow needs exactly these three: who is
+// asking (client_id), proof it is really them (client_secret), and the origin
+// the authorization code was issued against (redirect_uri).
+const (
+	FieldClientID     = "client_id"
+	FieldClientSecret = "client_secret"
+	FieldRedirectURI  = "redirect_uri"
+)
+
+// Fields is the full registration, in the order a form should render it.
+var Fields = []string{FieldClientID, FieldClientSecret, FieldRedirectURI}
+
+// keyPrefix namespaces the app registration away from per-project credentials.
+// The registration is instance-scoped — one OAuth client per deployment per
+// provider, shared by every consumer and every project.
+const keyPrefix = "oauth-app"
+
+// Key returns the secret-provider key holding one field of a provider's
+// instance-scoped OAuth app registration — e.g. oauth-app-google-client-id.
+//
+// Lowercase alphanumerics and hyphens only. Cloud secret backends compose their
+// provider-side secret name straight from this key and restrict that name's
+// charset: GCP allows [A-Za-z0-9_-], Azure Key Vault only [A-Za-z0-9-]. Azure is
+// the strictest, so its alphabet is the one used here.
+//
+// The substitution is lossy — two ids differing only in their separators produce
+// the same key. That is safe here and would not be for arbitrary input: both
+// arguments come from a closed set of compile-time constants (this package's
+// three fields, and the provider ids declared in registry metadata), so a
+// collision is a thing a reviewer can see rather than a thing a customer can
+// cause.
+func Key(provider, field string) string {
+	return keyPrefix + "-" + safe(provider) + "-" + safe(field)
+}
+
+// ConfigKey returns the provider-config key an app-registration field is passed
+// to a factory under.
+//
+// It is namespaced away from a provider's own config fields deliberately:
+// "client_id" is a plausible name for something a provider asks the operator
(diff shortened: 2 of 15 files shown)


jale-dbox:
**Review rounds 2–3 (Codex, xhigh).** Round 2 was clean. Round 3 covered the revocation endpoint added since, and found one thing.

**[P2] Strip project-supplied OAuth app fields — valid, fixed.**

> For an `authorization_code` datasource, a project document can therefore supply or retain OAuth client fields even though the registration is meant to come only from deployment secrets, masking missing app registration or using stale per-project values.

Right, and it is the failure the separation exists to make impossible: a value left standing because its secret happened to be unset would hide a missing app registration behind a stale per-project one, and the connection would authenticate against an application nobody registered.

The namespace is now cleared first and unconditionally — for every auth method, not only three-legged ones — and only values read from the secret store are put back. The enterprise side stops these keys being *written* ([enterprise#294](https://github.com/decisionbox-io/decisionbox-enterprise/pull/294)); this stops any that are already there being *read*, which is the half that matters for a document written before the namespace was reserved.

Two tests, both mutation-checked. `make build`, `make test-go`, `make lint-go` clean.

Also since round 2: `AuthorizationCode` gained an optional `revoke_url` (`bc3abef`). Forgetting a refresh token is not the same as ending the authorization behind it, and without somewhere to send the revocation a datasource that is re-authorized, removed, or switched to another method leaves a live grant in the customer's provider account with nothing left able to end it.

— Co-coded with Jale 🤖

jale-dbox:
**Review round 4 (Codex, xhigh) — clean.**

> The changes add metadata and plumbing for authorization-code warehouse auth without breaking existing static auth behavior. Targeted Go and dashboard tests/type-checking pass, and I did not find a discrete introduced bug that is actionable within this patch.

No findings, and the log carries no interruption marker — a Codex run that is cut short still exits 0 and prints a review-shaped tail, so that is worth checking before calling a round clean. Two earlier rounds on the companion PR were void for exactly that reason.

That is a clean round on the final diff (`3de7ed8`), which includes everything the three previous rounds turned up: the test-slug collision under `-count=2`, the community form rendering an empty credential box for a flow with no fields, the `AuthMethodByID("")` guess, `revoke_url`, and the namespace clearing.

No Copilot review has been posted on this PR — it is not active on this repo — so there is no second pass to close.

This PR is the one to merge **first**: [enterprise#294](https://github.com/decisionbox-io/decisionbox-enterprise/pull/294) cannot compile until the registry change is on `feat/non-sql-datasources`.

— Co-coded with Jale 🤖

jale-dbox:
## Done — what, how, and what it cost

**What.** The community half of user-delegated OAuth for data sources (enterprise decisionbox-io/decisionbox-enterprise#292): what the registry has to be able to *say*, and what the agent has to be able to *read*, so the enterprise side can hold a grant on a connection.

**How.**

- **A three-legged auth method.** `AuthMethod` could only describe a form to type into; it gains a flow and an `AuthorizationCode` block carrying the endpoints, the revocation URL, and the scopes the consent must cover — the last because granular consent means the exchange has to check the grant *against* the list, not merely ask for it.
- **The app registration, keyed by OAuth provider.** It was keyed by datasource slug, which quietly asserts each consumer needs its own client. A customer who registers one Google client should use it everywhere; the key scheme moves to `libs/go-common/oauthreg`, which neither registry owns.
- **A credential a datasource can name.** `initWarehouseProvider` derived the secret key from the datasource id, which cannot reach a credential several datasources share. `credential_ref` names a slot; unset — every datasource that exists today — changes nothing.

## How it was verified

- `libs/go-common`, `services/agent` and `services/api` suites green; `golangci-lint` clean on all three; composed dashboard suite (739 tests) green.
- The credential-key change is pinned by a **characterization test written before it**, so "the derived path still works" was already passing rather than written to agree with the new code.
- Mutation-checked: ignoring the ref, leaving it in the provider config, accepting an unusable one, and treating an empty one as set are all caught.
- One pre-existing environmental failure in this sandbox (`TestNewClient_UnreachableHost` — the host resolves here), unrelated to this branch.

## Review rounds

**5 completed rounds, r5 clean.** 7 findings; 5 acted on. Two were security-relevant and neither was theoretical:

- **`credential_ref` could dereference any project secret.** I reasoned that a project-scoped read only reaches secrets the project already owns and stopped there — but those include the project's LLM and embedding credentials, admin-written and masked everywhere the API returns them. The ref is composed *into* a key now, in a namespace nothing else writes.
- **A shared credential was readable across providers.** Namespacing it away from other features left the credentials inside that namespace reachable by each other, and the Postgres provider reads `credentials_json` as its password with `host` from config — so a Postgres datasource naming an analytics connection's slot and an attacker's host would send that connection's Google refresh token to it. I checked the provider source rather than take the finding on trust. The key is bound to the consumer the credential was obtained for.

Also fixed: an unresolvable ref fell through to an empty credential, which BigQuery reads as "use application-default credentials" — the agent's own service account; the Azure secret-name limit applies to the composed `<namespace>-<projectID>-<key>`, not the key alone; and an inferred auth method was applied without being recorded, so this function and the provider factory disagreed about how a datasource authenticates.

One P2 scoped out: a datasource can be enqueued for indexing before its grant exists. `ProjectsHandler.Create` does set `pending_indexing`, but it also calls `rejectUnanchoredProject`, and the only provider declaring the flow is `CanAnchor: false` — so it needs a future provider that both anchors and signs in. The underlying "indexed before credentialled" behaviour predates this branch and applies to every auth method.

## Known limitation

Nothing reads `oauth-app-*` until the enterprise side writes it, so this merges as inert plumbing — which is the point: **enterprise CI cannot compile its half until this lands on the bridge branch.**

— Co-coded with Jale 🤖

Issue #347, opened 2026-08-28
Discovery analysis fails with context-length 400: max_tokens not budgeted against the model's context window


jale-dbox:
## Problem

A discovery analysis area fails (run → **Partial**) with a 400 from the model:

```
channelperformance: bedrock/openai-compat: InvokeModel failed: operation error Bedrock Runtime:
InvokeModel, StatusCode: 400, ValidationException: "This model's maximum context length is 202752
tokens. However, you requested 64000 output tokens and your prompt contains at least 138753 input
tokens, for a total of at least 202753 tokens. Please reduce the length of the input prompt or the
number of requested output tokens."
```

`138,753 input + 64,000 output = 202,753` → **exactly 1 token over** the model's (GLM-5) 202,752 context. The area is lost and the run degrades to Partial.

## Root cause

The discovery **analysis path never budgets output against the model's context window.** It:
- requests a **fixed** `max_tokens` = `GetMaxOutputTokens(provider, model)` (`services/agent/internal/discovery/orchestrator.go:905`) → the openai-compat **`DefaultMaxOutputTokens: 64000`** (`providers/llm/openai/provider.go:87`, added in #338) for an uncatalogued model, and
- lets the input grow to its own query-results budget `AnalysisQueryResultsBudgetTokens = 200_000` (`analysis_step_picker.go:40`).

**Nothing enforces `input + output ≤ model_context`.** The `Budget` type that does exactly this math — `ModelMaxInput − ReservedOutput − ReservedSystem − SafetyMargin` (`libs/go-common/llm/budget.go`) — is wired into the **/ask** path only, **not** discovery analysis. And there is **no adaptive retry** on a context-overflow 400 (`internal/ai/retry.go` handles other errors, not this), so the area just fails.

## Why now (likely regression)

**#338** raised the uncatalogued openai-compat `DefaultMaxOutputTokens` to **64,000** (previously ~8,000).
- Before: `138,753 + 8,000 = 146,753 < 202,752` → fit.
- After: `138,753 + 64,000 = 202,753 > 202,752` → overflow.

So large-input analyses on ~200K-context models now overflow. Compounded by **GLM-5 being uncatalogued** — DecisionBox doesn't know its real 202,752 window, so it can't budget even if it wanted to.

## Fix

[...]

PR #404, opened 2026-09-10
fix(agent): budget reflection/questions output against the model cap, not a fixed default

### LINKED ISSUE #403: Reflection phase silently truncates: fixed 3000-token output default overrides the dynamic budgeting from #347/#348
## Symptom

On a project with a substantial Discovery Ledger, the reflection phase produces **findings but nothing else**: `coverage.explored_tables` is null, `coverage.summary` is empty, and there are zero learnings and zero next-tasks. The UI reads "0 of N tables explored" with a full frontier, so the feature looks broken rather than truncated.

Observed ledger document:

```
coverage: { explored_tables: null, area_depth: null, total_tables: 22, summary: '' }
convergence: [ { new_findings: 43, total_findings: 43, marginal_ratio: 1 } ]
learnings: 0, next_tasks: 0
```

## Cause

`RunPhaseReflection` has two halves. `consolidateFindings` (step 1) always runs — hence the 43 findings. Step 2 is a single bounded LLM call producing coverage, learnings, next-tasks and pack deltas. It failed, so `ref = nil`, and steps 3–4 still ran with nil. That is exactly the state above: everything non-LLM persisted, everything LLM-derived is empty.

Step 2 failed because the response was **truncated at max_tokens**. From the gateway's own usage records for the run:

| time | in | out | |
|---|---|---|---|
| 20:21:45 | 15337 | **3000** | attempt 1 — hit the cap exactly |
| 20:22:28 | 15461 | 2665 | attempt 2 (single repair retry) |

The +124 input tokens on the second call is `reflectionRepairSuffix(lastErr)` being appended, which only happens on the parse-failure path. The ledger's `updated_at` is 38 ms after the second call returned — the phase gave up and wrote findings-only.

3000 is `defaultDiscoveryReflectionMaxOutput`.

## Why the dynamic budgeting didn't help

#347/#348 made output budgeting dynamic. All four discovery LLM phases call the same `budgetedMaxOutputTokens(window, input, outputCap, floor)`. The difference is what each passes as `outputCap`:

| phase | outputCap source | binding limit |
|---|---|---|
| analysis (`orchestrator.go:1144`) | `resolveModelBudget()` → model cap | dynamic |
| recommendations (`orchestrator.go:1742`) | `resolveModelBudget()` → model cap | dynamic |
| clarifying questions (`questions.go:196`) | env, default **2000**, then min'd with model cap | fixed 2000 |
| reflection (`reflection_llm.go:84`) | env, default **3000**, then min'd with model cap | fixed 3000 |

The two phases that work take the model's cap directly. The two newer ones layer a small fixed default on top, and since the budget takes the minimum, that default wins and the dynamic sizing never applies.

The mismatch in scale: `defaultAnalysisMinOutputTokens` is **8192** — the *floor* every one of these calls is guaranteed. Reflection was *capped* at 3000, well under half the floor for a comparable call. Nothing upstream imposed this: the same model returned 13,399 output tokens on an analysis call in the same minute, and `DefaultMaxOutputTokens` for an uncatalogued model is 64000.

The reflection response is also the one that grows with usage — it re-judges every prior ledger finding, so the longer a project runs, the more certain this becomes. Questions has the identical latent bug at 2000; it currently works only because its output is small.

## Fix

PR #403.

- Default both phases to the model's own cap via a shared `phaseOutputCap` helper, matching the analysis/recommendation paths. `DISCOVERY_REFLECTION_MAX_OUTPUT` / `DISCOVERY_QUESTIONS_MAX_OUTPUT` remain operator overrides, still clamped and never above what the model accepts.
- An unknown model cap (`<= 0`) falls back to the old constant rather than 0 — `boundOutputCap` keeps a 0 cap at 0, which would collapse `max_tokens` and the floor to nothing.
- Log the case that was invisible: a response that fails to parse **and** consumed its entire budget is almost certainly truncation, which needs a different fix than malformed output. That distinction is the reason this went unnoticed.

## Not a bug

`total_tables: 22` is correct — `TotalTables = len(result.Schemas)`, the run's scoped working set, not the full warehouse catalog.

jale-dbox:
Closes #403.

## The bug

#347/#348 made output budgeting dynamic. All four discovery LLM phases call the same `budgetedMaxOutputTokens(window, input, outputCap, floor)` — the difference is what each passes as `outputCap`:

| phase | outputCap source | binding limit |
|---|---|---|
| analysis (`orchestrator.go:1144`) | `resolveModelBudget()` → model cap | dynamic |
| recommendations (`orchestrator.go:1742`) | `resolveModelBudget()` → model cap | dynamic |
| clarifying questions (`questions.go:196`) | env, default **2000**, then min'd with model cap | fixed 2000 |
| reflection (`reflection_llm.go:84`) | env, default **3000**, then min'd with model cap | fixed 3000 |

The two phases that work take the model's cap directly. The two newer ones layer a small fixed default on top — and since the budget takes the minimum, that default wins and the dynamic sizing never applies.

The scale mismatch: `defaultAnalysisMinOutputTokens` is **8192** — the *floor* every one of these calls is guaranteed. Reflection was *capped* at 3000, well under half the floor for a comparable call. Nothing upstream imposed it: the same model returned **13,399** output tokens on an analysis call in the same minute, and `DefaultMaxOutputTokens` for an uncatalogued model is 64000.

## What it looked like in production

A project with 43 prior ledger findings. The reflection response is sized by the ledger — it re-judges every prior finding — so it hit `max_tokens` exactly, truncated mid-JSON, failed to parse, and after the single repair retry `generateReflection` returned nil.

Steps 3 and 4 still run with `ref = nil`, so the ledger kept its findings and convergence row while `coverage.explored_tables`, `summary`, learnings and next-tasks all stayed empty. The phase looked like it had worked. From the gateway's usage records:

```
20:21:45  in=15337  out=3000   <- hit the cap exactly
20:22:28  in=15461  out=2665   <- repair retry (+124 input = the repair suffix)
ledger updated_at = 20:22:28.421   <- 38ms later, findings-only
```

This gets *more* likely over time, since the ledger only grows. Questions has the identical latent bug at 2000 — it works today only because its output is small.

## The change

- Shared `phaseOutputCap` helper: unset env → the model's own cap, matching analysis/recommendations. `DISCOVERY_REFLECTION_MAX_OUTPUT` / `DISCOVERY_QUESTIONS_MAX_OUTPUT` stay as operator overrides, still clamped and never above what the model accepts.
- An unknown model cap (`<= 0`) falls back to the old constant rather than 0. `boundOutputCap` keeps a 0 cap at 0, which would collapse `max_tokens` **and the floor** to nothing — worth guarding explicitly.
- `logOutputCapTruncation`: a response that fails to parse **and** consumed its entire budget is almost certainly truncation, not malformed output. Those need different fixes, and the absence of that distinction is precisely why this stayed hidden.
- Docs updated — both knobs now document "model's output cap" as the default and warn what setting them too low does.

## Tests

Six cases on `phaseOutputCap`, including a regression test that reproduces the production numbers (15337-token prompt, 200K window) and asserts the old path caps at exactly 3000 while the new one clears the 8192 analysis floor.

```
go test ./internal/discovery/ -count=1   ok (9.9s)
go vet ./internal/discovery/             clean
gofmt                                    clean on touched files
```

## Note on scope

This does not change analysis or recommendations — they were already correct. It brings the two newer phases in line with them.

https://claude.ai/code/session_014PDGPb3ae8nV1EeTMqEYCx

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -77,7 +77,7 @@ Generation is off by default and, where it is enabled, respects a per-project **
 |----------|---------|-------------|
 | `DISCOVERY_QUESTIONS_ENABLED` | `false` | Master switch for question generation. Off by default. The agent runs as a separate container, so this is forwarded from the API process to spawned agent containers automatically. |
 | `DISCOVERY_QUESTIONS_MAX` | `5` | Hard cap on the number of questions generated per run (after dedup against already-asked / already-answered questions). |
-| `DISCOVERY_QUESTIONS_MAX_OUTPUT` | `2000` | Output-token budget for the single generation call, budgeted against the model's context window like the analysis/recommendation phases. |
+| `DISCOVERY_QUESTIONS_MAX_OUTPUT` | model's output cap | Optional ceiling on the single generation call. Unset, the phase uses the model's own output cap and budgets it against the context window, like the analysis/recommendation phases. Set it only to deliberately cap the spend — a value below what the response needs truncates it mid-JSON and the phase yields nothing. |
 | `DISCOVERY_QUESTIONS_CONFIDENCE_MAX_PCT` | `50` | An insight/recommendation with confidence below this percentage is treated as "uncertain" and eligible to raise a question. |
 | `DISCOVERY_QUESTIONS_PARSE_MAX_RETRIES` | `1` | How many times to re-prompt if the model's response can't be parsed as the questions envelope. |
 | `DISCOVERY_QUESTIONS_TIMEOUT` | `3m` | Dedicated wall-clock budget for the questions hop, independent of `DISCOVERY_MAX_DURATION`. Go duration format. |
@@ -93,7 +93,7 @@ Semantic dedup / trend detection and per-analysis-area retrieval of prior findin
 |----------|---------|-------------|
 | `DISCOVERY_REFLECTION_ENABLED` | `false` | Master switch for the reflection / Discovery Ledger hop. Off by default. Forwarded from the API process to spawned agent containers automatically. |
 | `DISCOVERY_REFLECTION_TIMEOUT` | `3m` | Dedicated wall-clock budget for the reflection hop, independent of `DISCOVERY_MAX_DURATION`. Go duration format. |
-| `DISCOVERY_REFLECTION_MAX_OUTPUT` | `3000` | Output-token budget for the single reflection call, budgeted against the model's context window. |
+| `DISCOVERY_REFLECTION_MAX_OUTPUT` | model's output cap | Optional ceiling on the single reflection call. Unset, the phase uses the model's own output cap and budgets it against the context window. Set it only to deliberately cap the spend — the reflection response re-judges every prior ledger finding, so it grows with the ledger, and a value below what it needs truncates it mid-JSON and yields no coverage, learnings or next-tasks. |
 | `DISCOVERY_REFLECTION_PARSE_MAX_RETRIES` | `1` | How many times to re-prompt if the model's reflection response can't be parsed. |
 | `DISCOVERY_LEDGER_MAX_FINDINGS` | `500` | Retention cap on ledger findings per project; least-valuable (resolved/refuted, oldest) are pruned first. |
 | `DISCOVERY_LEDGER_DEDUP_MINSCORE` | `0.85` | Cosine-similarity floor above which a new finding is treated as a semantic duplicate of a prior one (needs the embedding provider). |
--- a/services/agent/internal/discovery/analysis_budget.go
+++ b/services/agent/internal/discovery/analysis_budget.go
@@ -2,10 +2,14 @@ package discovery
 
 import (
 	"context"
+	"os"
+	"strconv"
+	"strings"
 	"unicode/utf8"
 
 	goconfig "github.com/decisionbox-io/decisionbox/libs/go-common/config"
 	gollm "github.com/decisionbox-io/decisionbox/libs/go-common/llm"
+	applog "github.com/decisionbox-io/decisionbox/services/agent/internal/log"
 )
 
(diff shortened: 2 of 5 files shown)


abacigil (dismissed):

jale-dbox:
**Correction to the description** (raised in review — the original claim was an assumption, and the code contradicts it).

The description said the reflection response "gets *more* likely over time, since the ledger only grows". That is **wrong**. Every input to the reflection prompt is capped:

| input | cap |
|---|---|
| prior findings | **60** — `maxPriorFindingsInPrompt` |
| open tasks | **40** — `maxLedgerTasksInPrompt` |
| catalog tables | **300** — `renderCatalogTables` |

And the ledger self-manages: `renderOpenTasks` is fed `taskRepo.List(..., LedgerTaskStatusOpen)`, so tasks the ledger has closed or dropped fall out of the prompt entirely. Prior findings are sorted most-recently-seen first before the 60 cap, so the freshest survive; `Prune` keeps storage at `defaultDiscoveryLedgerMaxFindings` (500), but only 60 ever reach the prompt.

The output fields mirror those bounded inputs — `prior_status_updates` cannot exceed the 60 findings shown, `task_status_updates` cannot exceed the 40 tasks shown, `covered_tables` cannot exceed the catalog cap.

**So this is a flat undersize, not a scaling problem.** The observed failure was at 43 findings, comfortably inside the 60 cap: 3000 was simply below what an ordinary ledger needs, well within the design's own ceilings. It fails consistently once a project passes a modest threshold, rather than degrading progressively.

The fix is unchanged and still correct — the model cap is the right ceiling, and these bounded inputs are exactly what keep that safe rather than open-ended. Code comment corrected in the latest commit.

abacigil (approved):

PR #406, opened 2026-09-10
fix(ui): stop unbounded /projects/{id}/status poll loop on the project page

### LINKED ISSUE #405: Project page polls /projects/{id}/status in an unbounded render loop, not on the 2s interval
## The problem

Opening a project page starts an unbounded request loop against `GET /api/v1/projects/{id}/status`. Not the 2-second poll — that one is gated correctly. This one runs as fast as the API answers, for as long as the tab is open, and it starts on load whether or not a run is in progress.

`ui/dashboard/src/app/projects/[id]/page.tsx`:

```ts
const pollStatus = useCallback(async () => {
  try {
    const status = await api.getProjectStatus(id);
    if (status?.run) {
      const newRun = status.run as unknown as DiscoveryRunStatus;
      if (dismissedRunId.current === newRun.id) return;
      ...
      setRun(newRun);              // unconditional
      ...
    }
  } catch { /* ignore */ }
}, [id, run]);                     // depends on run

useEffect(() => { pollStatus(); }, [pollStatus]);   // fires on the callback's identity
```

The cycle:

1. `pollStatus()` fetches and calls `setRun(newRun)`.
2. `newRun` is freshly parsed JSON, so it is **always a new object** — `Object.is` never matches and React never bails out of the update.
3. `run` changes identity, so `useCallback` produces a new `pollStatus`.
4. `useEffect(..., [pollStatus])` sees a new dependency and calls it again.
5. Go to 1.

The only exits are a project with no run at all (`status.run` falsy) or a run whose id matches `dismissedRunId.current` — which returns *before* `setRun`. That second one is why dismissing a run appears to "fix" it.

## Impact

One request storm per open project tab, each hitting Mongo, for the entire time the tab is open. It is invisible in the API logs because only errors are logged, which is presumably why it has gone unnoticed.

## The fix, and the half that is easy to miss

Two changes, and both are needed:

1. **Make `setRun` conditional.** Update only when something rendered actually changed — the run's id and status — instead of on every response. This is the root cause: an unconditional `setState` of a fresh object inside a callback that is a dependency of the effect that calls it.
2. **Key the initial fetch on `id`, not on the callback's identity.** `useEffect(() => { pollStatus(); }, [pollStatus])` means "re-run whenever this function is recreated", which is never what was wanted — it wanted "once per project".

Doing only (1) removes today's loop but leaves the structure intact: the next unconditional `setState` reachable from `pollStatus` brings it straight back. Doing only (2) leaves the 2-second interval effect re-subscribing on every response, since it also depends on `pollStatus`.

The 2-second interval itself is fine and should stay as it is — it is correctly gated on `running`/`pending`.

## Overlay

`decisionbox-enterprise/ui/src/app/projects/[id]/page.tsx` overlays this file and the affected block is **byte-identical**, so the enterprise copy has to be fixed in the same change or the enterprise build keeps the loop.

Observed on a running stack: the endpoint was being called continuously with no discovery run active.

— Co-coded with Jale 🤖

jale-dbox:
## What

Opening a project page started an **unbounded** request loop against `GET /api/v1/projects/{id}/status` — not the 2-second poll (that one is correctly gated), but a separate storm that ran as fast as the API answered, for as long as the tab was open, whether or not a run was in progress. This stops it.

## Why

Only errors are logged for that endpoint, so the storm was invisible in the API logs while quietly hitting Mongo once per open project tab. The root cause was a React feedback loop:

1. `pollStatus()` called `setRun(newRun)` unconditionally.
2. `newRun` is freshly-parsed JSON, so it is **always a new object** — `Object.is` never matches and React never bails out of the update.
3. `run` changing identity re-created the `pollStatus` `useCallback`.
4. `useEffect(() => { pollStatus(); }, [pollStatus])` saw a new dependency and fired again → back to 1.

The only exits were "no run at all" or "run id already dismissed" (which returned *before* `setRun`), which is why dismissing a run appeared to "fix" it.

## How

Two changes, both required — fixing only one leaves the structure primed to loop again:

1. **`setRun` is now conditional** — the run is replaced only when a rendered field actually changed (`id` / `status` / `updated_at`). `updated_at` is bumped server-side on every write to the run doc, so the live progress header (progress %, query/insight counts) keeps ticking during a run, while the no-op responses that drove the loop are skipped.
2. **The load-time fetch is keyed on the project `id`**, not on the callback's identity — so it runs once per project ("once on load"), which is what it always wanted.

The 2-second interval effect is unchanged — it was already correctly gated on `running` / `pending`.

## Verification

- New Jest regression test (`ui/dashboard/src/__tests__/ProjectPage.test.tsx`): asserts a single status fetch on load when a terminal run is present, no fetch when there is no run, and that the 2s interval still polls, updates the live header, fires the running→done side-effects, then stops once the run is terminal. Confirmed the test **fails on the pre-fix code** (2 / 4 status calls instead of 1) and passes on the fix.
- Local: `make lint-ui` (0 errors), `make test-ui` (388 passing, incl. the 3 new), dashboard `npm run build` — all green.

Closes #405

— Co-coded with Jale 🤖


jale-dbox:
## Final report

**What** — Opening a project page started an *unbounded* request loop against `GET /api/v1/projects/{id}/status` — separate from the correctly-gated 2-second poll — that ran as fast as the API answered for the whole time the tab was open, whenever the project had a run (running **or** already finished). This eliminates it.

**Why** — A React feedback loop: `pollStatus()` called `setRun(newRun)` unconditionally; `newRun` is freshly-parsed JSON (always a new object, so `Object.is` never matches and React never bails out); `run` changing identity re-created the `pollStatus` `useCallback`; and `useEffect(() => { pollStatus(); }, [pollStatus])` re-fired on the new identity → back to the top. Only errors are logged for that endpoint, so it was invisible in API logs while hitting Mongo once per open tab.

**How** —
1. **`setRun` is now conditional** — the run is replaced only when a rendered field actually changed (`id` / `status` / `updated_at`). `updated_at` is bumped server-side on every write to the run doc, so the live progress header keeps ticking during a run while the no-op responses that drove the loop are skipped.
2. **The load-time fetch is keyed on the project `id`** instead of the callback's identity, so it fires once per project.

The 2-second interval effect is unchanged — it was already correctly gated on `running` / `pending`.

**Verification** —
- New Jest suite `ProjectPage.test.tsx` (4 cases): a single status fetch on load with a terminal run, no fetch without a run, the interval keeps polling + updates the live header + fires the running→done side-effects then stops, and the unchanged-run **skip** path. Confirmed to **fail on the pre-fix code** (2 / 4 status calls instead of 1) and pass on the fix.
- Local: `make lint-ui` (0 errors), `make test-ui` (388 passing, incl. the 4 new), dashboard `npm run build` — all green.
- Review: **Codex** — 2 rounds, both zero findings; round 2 re-examined the latest commit. CI green on the latest commit including `codecov/patch` — the round-2 commit added the skip-branch test that was the one partial-coverage line. Copilot code review is not active on this repo (no recent PR here carries a Copilot review), so Codex was the reviewer.

**Notes** — The project-board add (`gh project item-add 4`) failed: the run's token lacks project write scope. Labels (`bug`, `component: dashboard`, `priority: P2`), milestone (`v1.0.0`), and assignee (`abacigil`) are set.

— Co-coded with Jale 🤖

abacigil (approved):

PR #408, opened 2026-09-10
feat(agent,api): enforce table filters at index time + pre-index table preview; no auto-index on create

jale-dbox:
## Summary

Enforce table filtering at the **schema-index** stage and let operators review a warehouse's tables before indexing. Two behaviour changes + one new capability:

1. **Registered `ListTables` filters now run on the schema-index pass, not only discovery.** `runIndexSchema` now activates the agent plugin registry (`gosources.Configure`, exactly as `runDiscovery` does), so any registered table filter has a live handle and shrinks the enumerated set **before** per-table schema pull, description, and embedding. The schema index — and everything that reads it — is built over the filtered set, not the full catalog. No-op in the community build (no filter registered).

2. **Projects no longer auto-index on creation.** Creating a project with a warehouse used to immediately kick off a full index (schema + description + embedding for *every* table). On a 1,000+ table warehouse that spends time and model tokens on tables you may never want. A new project now stays "not yet indexed" until you press **Build schema index** — so you can inspect the tables (and, via a filter plugin, narrow them) first. The dashboard already renders a start button for the un-indexed state.

3. **Preview a warehouse's tables before indexing.** `GET /api/v1/projects/{id}/schema-cache/tables` falls back to a live, names-only enumeration (a cheap `ListTables` metadata pass — no schema, descriptions, or embeddings) when the project isn't indexed yet, so the UI can list the tables up front. Backed by a new agent `--list-tables` mode; degrades to an empty list on failure.

## How it was verified

- `go build ./...` for the agent and api modules; `go vet` + `golangci-lint` clean on the touched packages.
- Unit tests: live-fallback behaviour of `ListCachedTables` (empty cache → live list; non-empty cache → skip; error → empty), `AgentTableLister` JSON parsing (success / agent-error / success=false / nil runner), and the create-does-not-auto-index handler tests (updated from the old auto-enqueue assertions).
- Docs: `docs/concepts/plugin-hooks.md` (the filter now governs the index pass) + `CHANGELOG.md`; `make lint-docs` clean.

## Notes

Companion change coordinated in a private downstream repo (the table-filter plugin that drives points 1–3 there); this PR is the platform-side enforcement + enumeration and is a no-op without a registered filter.

## Update — per-datasource table picker (multi-warehouse)

`GET /schema-cache/tables` now accepts `?warehouse_id=<id>`: empty resolves to the primary (unchanged single-warehouse behaviour); an explicit id must name a real warehouse of the project (else `404`, so a bad id can't spawn a live enumeration against an arbitrary datasource). The resolved id threads into the cached lookup, the live `--list-tables` fallback, and the per-datasource live-table cache key — so the dashboard can show each datasource's own tables on a multi-warehouse project. Tests: explicit id scopes the cache lookup, empty = primary, unknown id → 404 with no live spawn, per-datasource live cache key. Codex review: clean (round 1).

— Co-coded with Jale 🤖


jale-dbox:
## Final report

**What** — Enforce table filtering at the schema-index stage (not just discovery), stop auto-indexing on project creation, and let operators preview a warehouse's tables before indexing.

**How**
- `runIndexSchema` now activates the agent plugin registry (`gosources.Configure`, as `runDiscovery` does), so a registered `ListTables` filter shrinks the enumerated set **before** per-table schema pull / description / embedding. `BuildIndex` also applies the cached-schema filter on a cache **hit** (where `DiscoverSchemas` is skipped), rejecting invented keys like the discovery orchestrator does.
- Project creation no longer auto-enqueues indexing (and the obsolete startup backfill that re-enqueued empty-status projects was removed); a project stays "not yet indexed" until the operator presses **Build schema index**.
- `GET /schema-cache/tables` falls back to a live, names-only `--list-tables` agent run (new agent mode) when the cache is empty and a warehouse is configured, so the picker is populated pre-index. Bounded by a per-(project, datasource-config) TTL cache + singleflight; a configurable `LIST_TABLES_TIMEOUT_SECONDS` is honored by the agent, the API wait, and the K8s Job deadline (with startup headroom) and is forwarded to spawned agents.

**Why** — Supports enterprise #306: choosing a table scope up front constrains index → blurbs → embeddings → Ask → pack-gen → discovery from the start, instead of processing thousands of unwanted tables on large (1,000+ table) warehouses. No-op in the community build (no filter registered).

**Verified**
- `go build` / `go vet` / `golangci-lint` clean across agent + api; unit tests (live-fallback: empty→list, non-empty→skip, error→empty, no-warehouse→skip, TTL-coalesced; `AgentTableLister` parsing; create-does-not-auto-index; filter narrowing before introspection). `make lint-docs` clean.
- **Live E2E** against a real 1,464-table MSSQL warehouse scoped to 22 tables: `ListTables filter applied … before=1464 after=22`, exactly 22 tables introspected, `project_schema_cache` = 22 rows, zero out-of-scope — and the pre-index picker enumerated all 1,464 incl. the 22 targets.
- **Codex review: 11 fix-rounds**, each a real issue (startup re-enqueue defeating no-auto-index; cache-hit filter bypass; list-error false-success; filter-contract length-trust; create-status clobber; warehouse-guard; cache-hit gate; live-listing spawn-loop → singleflight → datasource cache-keying → timeout config across agent/API/K8s-Job/env-forwarding/Job-headroom). CI green throughout.

**Round 12 — WONTFIX (rationale).** The two remaining comments concern plugin shapes that don't exist: (1) configuring plugins before the `--list-tables` filter — the only registered `ListTables` filter is the user-editable discovery-scope, and configuring it here would make the picker scope-filter *itself*, so you couldn't add a table to an allow-list; (2) a `CachedSchema`-only plugin altering the physical index on a cache-hit rebuild — the only registrant (discovery-scope) registers **both** filters with identical scope, so cold (`ListTables`) and cached (`CachedSchema`) rebuilds produce the same scoped index (no cache-state dependence). Both are theoretical for the current system; revisit if a non-scope policy `ListTables` filter or a `CachedSchema`-only plugin is ever added.

**Follow-ups / limitations** — The pre-index live listing is synchronous (one agent run per datasource-config per TTL, coalesced); on a very large BigQuery catalog raise `LIST_TABLES_TIMEOUT_SECONDS` (and any proxy read-timeout) accordingly. An async enumerate-to-store variant is a possible future enhancement if needed.

Companion (enterprise): decisionbox-enterprise#307.

— Co-coded with Jale 🤖

jale-dbox:
### Multi-warehouse follow-on — platform side

**What:** `GET /api/v1/projects/{id}/schema-cache/tables` now honours `?warehouse_id=<id>` so the discovery-scope table picker can show each datasource's own tables on a multi-warehouse project.

**How:** `ListCachedTables` resolves the id (empty → primary, unchanged; explicit id validated via `WarehouseByID` → `404` if it isn't on the project, so a bad/foreign id can't spawn a live `--list-tables` enumeration against an arbitrary datasource) and threads it into the cached lookup, the live-listing fallback, and the per-datasource live-table TTL cache key. The agent's `--list-tables --warehouse-id` already resolved the warehouse, so no agent change was needed.

**Why:** enforcement was already per-(project, warehouse); the picker was the one endpoint still hardcoded to the primary (#306).

**Verified:** `go test ./internal/handler` + new cases (explicit id scopes the lookup, empty = primary, unknown id → 404 with no live spawn, per-datasource live cache key); `go vet` clean. Live E2E from the enterprise side (see #307) exercised this endpoint against a real MSSQL primary (1464 `dbo.*`) + a Postgres secondary (4 `public.*`), unknown id → 404.

**Review:** Codex review clean on round 1 (no actionable issues).

**Limitations:** none new here; the single-warehouse path is byte-identical (empty `warehouse_id`).

— Co-coded with Jale 🤖

abacigil (approved):

PR #409, opened 2026-09-10
feat(agent): ask-serve seed anchoring + supportive context + mutation-tool seam

jale-dbox:
Strengthens the agentic **data-query Ask** (`services/agent/internal/askserve`) loop. Generic, provider-shaped seams only — all **no-ops when nothing is registered**, so the community build is unchanged. The enterprise wiring (KB/notes provider, `save_note` tool, delegate + UI) lands in the paired `decisionbox-enterprise` PR; this is the platform half of that bridge pair.

## What changed

**Part A — seed anchoring.** A seeded Ask conversation ("Ask about this") now anchors on its insight/recommendation instead of answering globally:
- The `FOCUS` block is a **directive** — scope a quantitative/ambiguous question to the entity (its tables/metric/segment) unless the user explicitly broadens — keeping the anti-injection framing but dropping the "optional" reading.
- **Seed-primed retrieval**: on the first (empty-history) turn, the entity's related insights + tables are searched up front and injected as reference context, so the loop starts grounded on the entity (structural pull, not just prose). Bounded — runs once, doesn't consume the query budget.
- The **seed is always a citation** on the finalized answer.

**Part B — supportive context.**
- The datasource **card** (subject areas / key entities / key metrics) is now rendered for **single-warehouse** projects too (previously multi-warehouse only).
- A compact **`PROJECT CONTEXT`** block (the project's business summary + base-context) is injected — already-bounded fields, same as discovery.
- A new **`KnowledgeProvider`** backs a **`search_knowledge`** tool over the project's documents + notes. The knowledge-sources provider is now activated in **serve mode** (`gosources.Configure` in `runAskServe` — it was previously only called on the discovery path), and `sources.IsConfigured()` gates the tool so a community build (no-op provider) never offers it.

**Part C — mutation-tool seam.** A new generic `libs/go-common/askmutation` registry lets a plugin add a **write** tool to the read-only loop. The loop dispatches a registered mutation call, records its proposal id on the transcript, and — because a write is not evidence — it **never grounds a data answer**, but a successful write lets the model finish the turn to confirm it. When a mutation tool is present the system prompt stops claiming the agent is read-only.

## Design notes
- No new env vars / feature flags: each capability is offered when its provider/plugin is present (same "offered when present" rule `search_insights` uses).
- The `TurnRequest` gains only an optional `caller_sub` (proposal audit).
- Mutation tools are offered on the **native tool-calling** path (the path the chat uses); the JSON-text fallback keeps its fixed read-only vocabulary.

## Tests
- `seed_prime_test.go`, `prompt_seed_test.go`: priming runs on a seeded first turn (not follow-ups/unseeded), grounds, folds hits, seed cited; failed search not appended; end-to-end.
- `prompt_context_test.go`: card rendered single-warehouse; `PROJECT CONTEXT` present/absent.
- `knowledge_test.go`: `search_knowledge` offered only with a provider, grounds, parses (native + JSON-text), error paths, event-preview cap.
- `mutation_test.go` + `askmutation` registry tests: offered when registered, records proposal id, doesn't ground but unlocks answering, capability line, failure path.
- `make build` / `go test ./internal/askserve/... ./agentserver/...` / `golangci-lint` — green.

— Co-coded with Jale 🤖


jale-dbox:
## Final report — #309 ask-serve: seed anchoring + supportive context + note-write seam

**What** (three parts of #309, platform half of the bridge pair with decisionbox-enterprise#310):

- **A — seed anchoring.** A seeded Ask conversation now anchors on its insight/recommendation instead of answering globally: a directive FOCUS block scopes a quantitative/ambiguous question to that entity, the entity's related insights + tables are retrieved up front on the first turn (structural grounding, not just prose), the seed is fed into the datasource router, and the seed is always a citation on the answer.
- **B — supportive context.** A `search_knowledge` tool over the project's documents + notes (backed by a `KnowledgeProvider` seam wired from the go-common sources registry — which is now also activated in serve mode, previously discovery-only); a compact `PROJECT CONTEXT` block (business summary); and the datasource **card** rendered for single-warehouse projects too.
- **C — note-write seam.** A new `libs/go-common/askmutation` registry lets a plugin add a write tool. A mutation records an **approvable** change and lets the model confirm it, but is never evidence (it can't ground a data answer). The enterprise `save_note` plugin fills it (#310). Community build: no provider registered → the loop is byte-for-byte read-only.

Also fixes the **empty-reference bug** (debug task #9): `search_insights` scoped its vector search to `insight`/`recommendation` and drops hits it can't enrich (with overfetch so valid hits ranked below stale vectors aren't starved), so a stale `ledger_finding`/`source_chunk` vector no longer surfaces as an empty-titled citation `[2]`/`[4]`. (Debug task #8 — the saved note not appearing in sentia — is the pending-proposal card, which #310 renders on the data-query path; the proposal itself was created correctly.)

**How verified.**
- `go build ./...`, `go test ./internal/askserve/… ./internal/insightsearch/…`, `golangci-lint` all clean locally; new tests cover seed priming/anchoring, KB search (+ empty-is-non-grounding + injection fencing), the mutation seam, the router bypass, and the insight overfetch/drop.
- **CI green**: Go Build, Go Test, Go Lint, both image builds (decisionbox-agent, decisionbox-api), CodeQL (go + js).
- **E2E on the prod `sentia` tenant** (local, tunnelled Mongo/Qdrant + warehouse, Turkish session): seed-anchored answers, `search_knowledge`, and the save path all exercised on real data.

**Review.** Codex review ran ~23 rounds — an unusually long, self-contained hardening loop concentrated on the brand-new write-tool path and the prompt-injection surface. What it caught and I fixed, in order of theme:
- *Write integrity*: a mutation never grounds a data answer; a requested-but-deferred save is never silently dropped (tracking evolved bool → counter → **tool-name-keyed set** to end an edge-case class cleanly) and never falsely reported ("not saved" vs "partially saved" vs no-op); a save-only turn confirms its outcome on **every** terminal (answer/clarify/decline), including a completed no-op; plugin args are deep-copied so the persisted event stays the model's pristine input; fail-closed member+ gate.
- *Prompt-injection fencing*: every piece of untrusted stored text fed to the model is now `%q`-quoted under a "data, not instructions" preamble — knowledge passages **and** source names/types, the seed, the project summary, warehouse/datasource descriptions + card fields, and the seed-primed context block.
- *Routing*: a knowledge-only/save-only turn on a multi-datasource project is no longer dead-ended by the datasource router — the router emits a `project_level` intent signal and the bypass is gated on it **plus** an actually-callable project-level tool (KB on either loop; write tools only on the native-tool loop).

The final two Codex rounds could not complete due to a Codex **remote-compaction infra error (404)** on the now-large diff — not a code finding. I stopped there as converged: the last completed round was fully addressed, and the only standing item is the by-design note below.

**@abacigil — decision flagged (WONTFIX here, not a blocker):** ask-serve is an internal backend service that trusts its whole request from the API delegate (which derives the caller role server-side from the authenticated session). #309 adds the first *write* behind that boundary; Codex twice flagged that the role arrives as a request field. Writes only create *pending, human-approved* proposals (low blast radius), and adding transport auth (mTLS / signed internal request) to the ask-serve endpoint is a platform-wide trust-boundary change beyond #309's scope — so I did **not** change it. Say the word if you want it as a follow-up hardening task.

**Follow-up.** Merge this before enterprise#310, then flip that PR's `.platform-ref` back to `main`.

— Co-coded with Jale 🤖

abacigil (approved):

PR #410, opened 2026-09-11
Non-SQL datasources: a datasource declares its query language, shape and whether it can anchor (platform half of enterprise#252)

jale-dbox:
Lands the platform half of the non-SQL datasource epic (decisionbox-io/decisionbox-enterprise#252) on `main`. 29 commits, developed on `feat/non-sql-datasources` since July, with `main` merged in three times along the way — most recently this morning.

## What this is

The thesis: **"SQL dialect" becomes "query language"**. A datasource declares at registration what it is — its query language, its *shape*, and whether it can **anchor** a project — and the platform reads that declaration without opening a connection. Everything else follows from not assuming tables.

## What changes for an existing deployment

**Nothing, and that is the load-bearing claim.** Every SQL warehouse is `entities`-shaped and anchoring, which is exactly what the descriptor defaults to. The prompt surfaces, both tool definitions and the grounding nudge were diffed against the previous implementation to confirm a project whose datasources are all SQL renders **byte-for-byte** as before. No warehouse declares a quality caveat, so those paths are inert. Packs for SQL sources are generated from the same exemplars by the same validator.

## The work

| | |
|---|---|
| **W1** | Query-language seam + capability descriptor (#358, #359) |
| **W3** | Credential lifecycle — SA token minting and caching (#360), then user-delegated OAuth (#395) |
| **W4** | Catalog-shaped schema indexing and retrieval (#364) |
| **W5** | Join-key validation: declare the key, validate the declaration (#375, #376, #377) |
| **W6** | Anchoring — resolve, demote-only, refuse a project that cannot carry itself (#362, #368) |
| **W7** | Result-quality caveats through to the insight (#363, #365) |
| **W8/W9** | Shape on a pack; a table identifier asked only where tables exist (#380, #385) |
| **W11** | A cube-shaped reference pack (#386) |
| **W12** | Cube exploration stops on no-new-signal, not a step count (#382) |
| **W13** | Router ballot and verdict, anchoring refusals, what a result is not (#361, #372, #373) |

W2 was **rejected** and stays rejected: per-language scope enforcement needs a SQL parser, which loses to CTEs, subqueries and dynamic SQL. Scope rests on the credential and on warehouse-side controls.

## Five silent failures this closes

Each produced well-formed, confident output rather than an error — none would have arrived as a bug report:

1. **A thresholded result looked exactly like a complete one.** Caveats now travel to the insight, and the insight's label is derived from its cited steps rather than authored by the model — a model that simply omitted the caveat would otherwise produce a finding indistinguishable from a sound one.
2. **A tenant filter was "verified" by looking for the field name in the query text** — which a request-shaped query contains while filtering by nothing. It passed, ran property-wide, and reported as scoped (#375).
3. **The repair loop told a non-SQL source to "fix this SQL query"**, inside a ```sql fence, contradicting the source's own template printed above it — and then discarded any reply that lacked `SELECT` (#367).
4. **The router could not see a cube source at all**, so spurious confidence from unrelated table matches pushed turns past the clarify threshold (#364).
5. **A project holding only observation sources** reached the agent looking healthy and confidently restated what those sources already report (#368).

## Verification

`make build`, all 30 Go modules, `lint-go` (0 issues), `test-ui` (394 tests). The first adapter built on this seam lives in the enterprise repo and is **gated off by default**, so nothing here reaches a customer dropdown on merge.

Paired with decisionbox-io/decisionbox-enterprise — that side must merge **after** this one, since its `.platform-ref` pins this branch.

— Co-coded with Jale 🤖


jale-dbox:
Known gap to settle before this merges: #412.

The Discovery Ledger counts coverage in tables (`ledger.Coverage.TotalTables = len(result.Schemas)`), and a source on the catalog path contributes none — so a run that explores a cube-shaped datasource records zero coverage for it and tells the next run the frontier is empty. Reflection is default-on at both gate layers.

Nothing in this PR causes it. The ledger landed on `main` while the query-language seam was on this branch, so the two have never been in the same build — which is exactly why this merge is the moment it becomes reachable. Filing rather than fixing here: it is a design question about what coverage means for a source whose unit is not a table, and this PR is already large.

#412 also folds in two smaller instances of the same cause — `questions.md`'s table-only `linked_target` vocabulary, and the unconditional "scope your … SQL to this … its tables" line in `writeSeedSection`.

— Co-coded with Jale 🤖

abacigil (approved):

Issue #412, opened 2026-09-11
Discovery Ledger counts coverage in tables, so a cube-shaped datasource reads as fully explored after every run


jale-dbox:
The Discovery Ledger's world model is a set of **tables**. A cube-shaped datasource has none — it offers a catalog of metrics and dimensions — so a run that explores one records **zero coverage for it**, and the next run is told the project is fully explored.

This is not "the cube is missing from a report". It is the compounding premise inverted: the feature exists so each run builds on the last, and for a cube source each run starts from a world model that says there is nothing left to look at.

## Where it comes from

Two lines, both table-typed by construction:

```go
// reflection_llm.go
p = strings.ReplaceAll(p, "{{CATALOG_TABLES}}", renderCatalogTables(result.Schemas))

// reflection.go
ledger.Coverage.TotalTables = len(result.Schemas)
```

`result.Schemas` is `map[string]models.TableSchema`. A source that implements `warehouse.CatalogSource` contributes none — that is the whole point of the catalog path in the schema indexer. So the count and the rendered catalog can only ever describe the table-shaped datasources in the run.

`prompts/reflection.md` then asks the model to work strictly within that vocabulary:

- `## Warehouse catalog (all tables — pick which were covered vs. still frontier)`
- **covered_tables**: *"the fully-qualified tables (`dataset.table`) this run actually queried. Copy names verbatim from the catalog. Omit tables you did not touch."*
- frontier policy: *"breadth_first = tile untouched **tables**"*
- **next_tasks**: *"Ground each in the findings/coverage above."*

A model obeying that prompt **cannot** report work it did against a cube, because there is no name in the catalog to copy.

## What a run actually produces

A project with one table-shaped primary (5 tables) and one cube source (~470 catalog items), after a run that queried both:

```
Explored 5 of 5 catalog tables (0 still on the frontier).
```

That string is what the next run inherits, from `previous_context.go`:

```go
fmt.Fprintf(&sb, "Explored %d of %d catalog tables (%d still on the frontier). ", explored, cov.TotalTables, frontier)
```

Consequences, in order of how much they compound:

1. **The frontier is empty when it is not.** Everything reachable through the cube is neither covered nor frontier — it is not in the world model at all.
2. **`next_tasks` can never propose cube work**, because they must be grounded in the coverage above.
[...]

PR #415, opened 2026-09-11
fix(agent): make Discovery Ledger coverage shape-aware

jale-dbox:
The Discovery Ledger's world model was a set of **tables**. A cube-shaped datasource has none — its queryable surface is a catalog of metrics and dimensions — so a run that explored one recorded **zero coverage for it**, and the next run was told the project was fully explored.

Not a cube missing from a report. The compounding premise inverted: the feature exists so each run builds on the last, and for a cube source each run started from a world model saying there was nothing left to look at — cumulatively, since every run inherits the coverage summary the last one wrote.

Addresses #412.

## What changed

**The reflection prompt renders a cube catalog beside the warehouse one.** Without it there is no name for the model to copy for work it genuinely did, so no amount of good reasoning could produce a report. The section also says what a cube is *not* — no frontier to tile — which is the load-bearing half: the rest of that prompt is written around a frontier that shrinks as it is covered, and a list of 470 metrics read under that framing is either reported as all-covered or treated as a backlog to exhaust. Neither is true of a combinatorial surface.

**Coverage is stored per shape.** `ExploredCatalogItems` is a record of what has been sliced, deliberately *not* a second fraction — a ratio over a cube's catalog would trade one false completion signal for another. `TotalCatalogItems` is written whether or not the LLM call succeeded, because it is what tells the next run this project has a cube at all, and the reflection hop is explicitly allowed to fail.

**Self-reported coverage is checked against the catalog it claims to come from.** Both namespaces are bare strings, so a cube metric answered into `covered_tables` was indistinguishable from a table and landed in the explored set verbatim — pushing the count past the catalog size, which the frontier arithmetic then clamped to zero. Wrong in both directions at once. Incoming names only: what earlier runs recorded is never revoked, so a narrowed discovery scope or a dropped table cannot erase the record of work that really happened. Matching is case-insensitive and resolves to the catalog's own spelling, which also stops an upper-case warehouse from counting one table twice.

**The ledger page says the counts are the warehouse side only.** Once every table is covered the UI rendered `Frontier 0 — tables not yet explored`: accurate as a subtitle, and read by a human as "nothing left" on a project where most of what it can query has never been touched.

**A seeded ask turn is anchored in the shape it can reach.** The FOCUS block is written *before* the datasources section, so on a seeded conversation it is the first thing the model reads about scoping — and it said "scope your retrieval and SQL to this insight — its tables, metric, and segment". Two false statements to a turn whose source has no tables and rejects SQL. Ordering mitigated it; mitigation is not truth.

## Nothing changes for a run that reaches only tables

The reflection prompt, its repair suffix, and the coverage line all render **byte-identically**. The golden they are held to was captured by rendering the same fixture on this branch's parent commit, so a drift of one byte fails the build rather than quietly changing what every SQL-only project's reflection is asked to do.

## Verification

- `go test ./...` green across `services/agent` and `libs/go-common`; `golangci-lint` clean on the touched packages; `gofmt` clean on the touched files. Two failures in the full sweep are environmental and pre-existing — `TestNewClient_UnreachableHost` (the dev Mongo occupies :27099, so the "unreachable" host answers) and `TestLoad_MissingMongoDBURI` (passes with `env -u MONGODB_URI`), both unrelated to this diff.
- All 396 dashboard jest tests green, eslint clean.
- The enterprise module builds against the changed `go-common` (additive fields only).
- **Mutation-tested, 14/14 caught.** Every new assertion was proven to fail against a deliberately broken implementation: the cube section suppressed, the `covered_catalog_items` bullet withheld, the catalog heading drifted by a single space, the repair suffix naming the cube field unconditionally, reported names trusted instead of checked, the catalog match made case-sensitive, the name index left unsorted and its tie-break reversed, the cube clause dropped from the coverage line, the cube total made dependent on the LLM call, the UI block suppressed and its API field zeroed, and the seed anchor pinned to each shape in turn. One early survivor was a real weakness in the test rather than in the code — it compared two orderings to each other, which a last-writer-wins index satisfies as readily as a first-writer one; it now names the expected spelling and catches both.

## Two things I did not do, and why

**#412's claim that convergence is wrong is incorrect — mine, and I'll correct it on the issue.** The marginal-new ratio is computed over ledger *findings*, and `buildFindingCandidates` reads `result.Insights`, which carries cube-derived insights like any other. Convergence was never shape-blind. The coverage *reading* was, which is what this PR fixes.

**`prompts/questions.md` is left alone.** I listed its table-flavoured vocabulary in #412, and on inspection it is a soft edge rather than a defect: a clarifying question about a cube can already link to the `insight` it came from or to an `area`, both valid target types, and `QuestionTargetLink` renders nothing for a `table` target, so even a junk one is inert in the UI. Changing that prompt would alter question generation for every existing project on a guess that a model *might* be discouraged by the wording. That is not a trade worth making without evidence — unlike the reflection prompt, where reporting cube work was structurally impossible.

Part of #252

— Co-coded with Jale 🤖

### DIFF
--- a/libs/go-common/models/discovery_ledger.go
+++ b/libs/go-common/models/discovery_ledger.go
@@ -60,9 +60,28 @@ type LedgerCoverage struct {
 	// AreaDepth maps analysis-area id -> a coarse "runs that produced findings
 	// in this area" counter, so depth-first policy can chase the richest seam.
 	AreaDepth map[string]int `bson:"area_depth" json:"area_depth"`
-	// TotalTables is the size of the indexed catalog at last update, so a
-	// frontier count (TotalTables - len(ExploredTables)) can be shown.
+	// TotalTables is the number of tables in the indexed catalog at last
+	// update, so a frontier count (TotalTables - len(ExploredTables)) can be
+	// shown. It counts the project's TABLE-shaped datasources only: a
+	// cube-shaped one has no tables to count, and folding its catalog in here
+	// would present a combinatorial surface as a frontier to exhaust.
 	TotalTables int `bson:"total_tables" json:"total_tables"`
+	// ExploredCatalogItems are the catalog refs — metrics and dimensions — the
+	// agent has queried on cube-shaped datasources across runs.
+	//
+	// Held apart from ExploredTables rather than merged into it because the two
+	// are not the same kind of thing to a frontier. A table is tileable: cover
+	// it and it is done, which is what makes "N of M" meaningful. A cube's
+	// slices are combinatorial, so there is no M at which it is finished, and a
+	// ratio built over one would read as a completion percentage for something
+	// that never completes. This set is a record of what has already been
+	// sliced, not a numerator.
+	ExploredCatalogItems []string `bson:"explored_catalog_items,omitempty" json:"explored_catalog_items,omitempty"`
+	// TotalCatalogItems is how many catalog items the run's cube-shaped
+	// datasources offer. Non-zero is also the signal that this project HAS a
+	// cube at all, which is what stops the coverage line from reporting an
+	// empty frontier once every table is covered.
+	TotalCatalogItems int `bson:"total_catalog_items,omitempty" json:"total_catalog_items,omitempty"`
 	// Summary is a short natural-language coverage note the reflection phase
 	// maintains ("orders + customers well covered; the events tables untouched").
 	Summary string `bson:"summary" json:"summary"`
--- a/services/agent/internal/askserve/prompt.go
+++ b/services/agent/internal/askserve/prompt.go
@@ -27,7 +27,7 @@ func buildSystemPrompt(rt *ProjectRuntime, routing turnRouting, cfg Config, char
 		b.WriteString("You are a data analyst agent. Answer the user's natural-language question about their data by reasoning step by step and running read-only SQL against their data warehouse. Ground every claim in query results — never invent numbers.\n\n")
 	}
 
-	writeSeedSection(&b, seed)
+	writeSeedSection(&b, seed, shapes)
 	writeDataSection(&b, routing)
 	writeProjectContextSection(&b, rt)
 
@@ -119,7 +119,7 @@ func buildSystemPromptForTools(rt *ProjectRuntime, routing turnRouting, cfg Conf
 		b.WriteString("You are a data analyst agent. Answer the user's natural-language question about their data by reasoning step by step and using the provided tools to run read-only SQL against their data warehouse. Ground every claim in query results — never invent numbers, table names, or column names.\n\n")
 	}
 
-	writeSeedSection(&b, seed)
+	writeSeedSection(&b, seed, shapes)
 	writeDataSection(&b, routing)
 	writeProjectContextSection(&b, rt)
 
@@ -234,7 +234,7 @@ const seedPromptTextCap = 800
 // turn on that entity — a quantitative or ambiguous question is scoped to the
 // entity rather than answered globally — while still deferring to an explicit
 // request to broaden. No-op when the turn is not seeded.
-func writeSeedSection(b *strings.Builder, seed *SeedContext) {
+func writeSeedSection(b *strings.Builder, seed *SeedContext, shapes sourceShapes) {
 	if seed == nil {
 		return
 	}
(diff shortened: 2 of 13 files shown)


jale-dbox:
## Final report

**What.** The Discovery Ledger's world model was a set of tables. A cube-shaped datasource has none, so a run that explored one recorded zero coverage for it and the next run read `Explored 5 of 5 catalog tables (0 still on the frontier)`. Not a cube missing from a report — the compounding premise inverted, and cumulatively, since every run inherits the summary the last one wrote. Fixed in the reflection prompt, the ledger write, the read context the next run gets, the ledger page, and the seeded-ask FOCUS block. 6 commits, `146cc4a0`.

**Why, tied to #412.** Three failures followed from the one cause: the frontier read as empty when it was not; `next_tasks` could never propose cube work, because they must be grounded in the coverage above; and both compounded run over run. The catalog rendered into the reflection prompt came from `result.Schemas`, which is `map[string]models.TableSchema` — a `warehouse.CatalogSource` contributes nothing to it by construction — so a model told to "copy names verbatim from the catalog" had no name to copy for work it genuinely did.

**How.**

- **`renderCatalogSection`** renders a cube catalog beside the warehouse one, and states what a cube is *not*: no frontier to tile. That half is load-bearing. The rest of that prompt is built around a frontier that shrinks as it is covered, and a list of 470 metrics read under that framing is either reported as all-covered or treated as a backlog to exhaust — neither true of a combinatorial surface.
- **Coverage is stored per shape.** `ExploredCatalogItems` is a record of what has been sliced, deliberately *not* a second fraction. `TotalCatalogItems` is written whether or not the reflection LLM succeeded, because it is what tells the next run the project has a cube at all, and that hop is explicitly allowed to fail.
- **`mergeExplored` checks self-reported names against the catalog they claim to come from.** Both namespaces are bare strings, so a cube metric answered into `covered_tables` was indistinguishable from a table and landed in the explored set verbatim — pushing the count past the catalog size, which the frontier arithmetic then clamped to zero. Wrong in both directions at once. Incoming names only: what earlier runs recorded is never revoked, so a narrowed discovery scope cannot erase work that really happened. Matching is case-insensitive and resolves to the catalog's own spelling, which also stops an upper-case warehouse from counting one table twice.
- **The ledger page** said `Frontier 0 — tables not yet explored` on the same premise. Accurate as a subtitle; read by a human as "nothing left".
- **`writeSeedSection`** told a seeded turn to "scope your retrieval and SQL to this insight — its tables, metric, and segment". It is written *before* the datasources section, so it is the first thing the model reads about scoping, and both halves are false for a source with no tables that rejects SQL.

A run that reaches only tables renders **byte-identically** — prompt, repair suffix, and coverage line. The golden was captured by rendering the same fixture on this branch's parent commit.

**Verification.**

- `go test ./...` green across `services/agent` and `libs/go-common`; `golangci-lint` clean on the touched packages; `gofmt` clean on the touched files. 396 dashboard jest tests green, eslint clean. CI green on `146cc4a0` — Go Build, Go Test, Go Lint, Dashboard Build, Dashboard Test & Lint, License Check.
- Two failures in the local full sweep are environmental and pre-existing, verified as such rather than assumed: `TestNewClient_UnreachableHost` (the dev MongoDB occupies :27099, so the "unreachable" host answers) and `TestLoad_MissingMongoDBURI` (passes under `env -u MONGODB_URI`). Codex independently reached the same conclusion in round 3.
- The enterprise module builds against the changed `go-common`; the additions are new fields only, and its ledger handler serialises the struct wholesale, so nothing there needed changing.
- **Mutation-tested, 18/18 caught.** Every assertion was proven to fail against a deliberately broken implementation — cube section suppressed, `covered_catalog_items` bullet withheld, catalog heading drifted by one space, repair suffix naming the cube field unconditionally, reported names trusted instead of checked, catalog match made case-sensitive, name index left unsorted and its tie-break reversed, cube clause dropped from the coverage line, cube total made dependent on the LLM call, `hasLedger` ignoring the cube total, the UI block suppressed, its API field zeroed, its denominator forced to print zero, and the seed anchor pinned to each shape in turn. One early survivor was a weakness in the test rather than the code: it compared two orderings to each other, which a last-writer-wins index satisfies as readily as a first-writer one. It now names the expected spelling and catches both.

**Review rounds.** Codex ×3. Round 1 raised two, both valid, both fixed in `65b0b95c`:

- *P2* — `hasLedger` ignored `TotalCatalogItems`, so a ledger holding only the cube's catalog size was discarded on load. That is precisely the state a run records when reflection produces nothing, which made the unconditional write a write for a read that never happened, and `renderCoverage`'s cube-total branch unreachable.
- *P3* — the dashboard tested only the total where the agent-side renderer tests either half, so a run whose cube catalog could not be read hid the slices earlier runs had recorded.

Round 2: *"I did not identify any discrete, actionable bugs introduced by this patch."* I then pushed one more commit — a stale comment saying "the 300-table catalog cap" when there are now two capped lists and the number lives in `maxCatalogNamesInPrompt` — and ran round 3 over it, since a clean verdict only covers the diff it saw. Round 3: *"No actionable correctness issues were found in the changes."*

Copilot has no review on this PR and none on #409, #410 or #411 — auto-review is not active on this repo, so there was no pass to close. Zero review threads outstanding.

**Two corrections to #412, both to my own claims, posted on the issue.**

1. **"Convergence is wrong" is not true.** The marginal-new ratio is computed over ledger *findings*: `consolidateFindings` builds candidates from `result.Insights` via `buildFindingCandidates`, which requires only a non-empty name. Nothing in that path reads a shape, and a cube-derived insight is carried like any other. The coverage *reading* was shape-blind; convergence never was.
2. **`prompts/questions.md` is a soft edge, not a defect, and is left alone.** A clarifying question about a cube can already link to the `insight` it came from or to an `area` — both valid target types — and `QuestionTargetLink` renders nothing for a `table` target, so even an invented one is inert in the UI. Changing that prompt would alter question generation for every existing project on a guess that a model *might* be discouraged by table-flavoured wording.

- I diverged from the fix sketch in #412, which suggested a per-datasource `explored N of M`. This is per-*shape*, and the cube deliberately has no denominator: a fraction over a combinatorial catalog is the same false completion signal in a new coordinate system. If per-datasource breakdown is wanted later it is an additive change on top of these fields.
- `mergeExplored` validates against what *exists* (the full `result.Schemas` map, the full catalog) rather than what the model was *shown* (each list is capped at 300 names). Deliberate: a table past the cap is still a real table, and dropping it would be a new way of losing coverage.
- Only incoming names are validated. A ledger that already carries junk from before this change keeps it; the frontier clamp handles the arithmetic. Self-healing would mean revoking stored coverage, which loses real history when a scope narrows.
- Nothing here changes `AGENT_FORWARD_ENV` or the gating work — this is the seam's correctness, not its rollout.

Part of #252

— Co-coded with Jale 🤖

PR #416, opened 2026-09-11
fix(ui): treat an API 401 as a re-authentication event, not a data error

jale-dbox:
## Summary

When an API call returns **401**, the dashboard client (`ui/dashboard/src/lib/api.ts` `request()`) threw a generic `ApiError` — exactly like any other non-2xx. Pages catch that into a toast and leave the entity `null`, so they render their empty state (e.g. `projects/[id]/page.tsx` → "Project not found"). The net effect: when a session expires, the user is stranded on a broken-looking data screen with no path back to sign-in, and refreshing reproduces it.

A 401 is not a data error — it means *this request is no longer authenticated*. The dashboard middleware is the single authority on where an unauthenticated user should go (a login page / SSO handshake / etc.), so on a 401 the client now hands the decision back to it: it reloads the current navigation so the middleware re-gates it and redirects. A `fetch`/XHR cannot itself redirect the browser, so this is the only way an in-page request that expires *mid-session* gets the user back to sign-in.

## What changed

- `request()`: on `res.status === 401`, call `handleUnauthorized()` and, if it schedules a re-auth navigation, return a never-settling promise so the caller doesn't briefly render a data error before the reload takes effect. Non-401 failures throw `ApiError` exactly as before.
- `handleUnauthorized()`: reloads the current page, **guarded** so that
  - concurrent 401s (a page issues several loads at once) trigger **at most one** reload,
  - a backend that keeps 401ing after a reload does **not** loop — after a 10s window it falls through to the normal error path,
  - the guard is cleared on any successful response so a later expiry re-triggers,
  - SSR / non-browser contexts fall through to the normal throw.
- Small `_reauth.navigate` seam so this is unit-testable (jsdom's `window.location` is non-configurable and its methods throw "Not implemented", so it can't be spied directly).

## Tests

`ui/dashboard/src/__tests__/api.test.ts` — 7 new cases: reloads once on a 401 and doesn't reject the caller; at most one reload under concurrent 401s; no reload (and normal error) when a 401 recurs inside the guard window; reload again after the window elapses; guard cleared on a 2xx; non-401 failures still throw `ApiError`. Full suite: 88 passed. `eslint` clean.


jale-dbox:
## Final report

**What** — `ui/dashboard/src/lib/api.ts` `request()` now treats a `401` as an authentication event instead of a generic error. On a 401 it hands off to the dashboard middleware (the authority on where an unauthenticated user goes) by re-running the current navigation, rather than throwing an `ApiError` that pages render as an empty "…not found" state.

**Why** — a `401` is "not signed in", not "no data". Surfacing it as a data error stranded a user with an expired session on a broken screen with no path back to sign-in, reproducible on every refresh. A `fetch`/XHR can't redirect the browser itself, so the client must react — this is the only way an in-page request that expires mid-session routes the user back to re-auth.

**How** — `handleUnauthorized()` reloads once, guarded so: concurrent 401s trigger a single reload; a persistent 401 after reload can't loop (a 10s sessionStorage-backed guard falls through to the normal error path); the guard clears on any 2xx; SSR/non-browser and storage-unavailable contexts fall through to the normal throw (no blind reload). A tiny `_reauth.navigate` seam keeps it unit-testable (jsdom can't stub `window.location`).

**Verified** —
- Unit: `api.test.ts` +8 cases (reload once on 401 & caller doesn't reject; single reload under concurrent 401s; no-loop within the guard window; reload again after it elapses; guard cleared on 2xx; non-401 still throws; storage-unavailable → throws, no navigation). Full suite **89 passed**; `eslint` clean; CI green.
- Review: Codex CLI, 2 rounds. Round 1 caught a reload-loop risk when `sessionStorage` is unavailable (fixed — write the guard before navigating, else fall through). Round 2 raised a point that only applies to a deployment whose middleware doesn't gate page navigations; in that configuration a 401 is not an expected auth outcome and correctly surfaces as an error, so no change (the reload path is exercised by deployments that do gate navigations).

**Limitations** — the reload delegates the destination to the middleware; a deployment with no page-gating middleware simply surfaces the error (there's nothing to route to), which is the correct behavior there.

— Co-coded with Jale 🤖

abacigil (approved):

PR #420, opened 2026-09-16
Automate GPG-signed release tags + gate unsigned tags

jale-dbox:
Automates signed release tags so signing can't be skipped by hand, and gates image publishes on a signed tag.

## What
1. **`cut-release.yml`** — a `Cut Signed Release` workflow_dispatch (input: `version`) that imports the release GPG key from secrets, confirms it matches the committed `release-signing-key.asc`, runs `git tag -s` + `git tag -v`, and pushes. The tag is born signed in CI — run it instead of `git tag` / `gh release create`.
2. **verify-tag gate** in `docker-publish.yml` — on a `v*` tag the publish requires the tag to be a valid GPG-signed tag; a lightweight/unsigned tag fails → no images ship. Branch/dispatch/PR builds pass (no tag).

## Needs
- Org secrets `GPG_PRIVATE_KEY` + `GPG_PASSPHRASE`.

## Verification
- GPG signing + `git tag -v` mechanics proven locally (signs non-interactively via a loopback-passphrase wrapper; verifies exit 0 on a good signed tag with only the public key, exit 1 on a lightweight tag).
- `actionlint` + shellcheck clean.

— Co-coded with Jale 🤖


jale-dbox:
Codex flagged that the verify-tag gate can be bypassed by pushing a `v*` tag at an **old commit** whose `docker-publish.yml` predates the gate (GitHub runs a tag-triggered workflow from the tagged commit's tree). This is a real limitation, but it is **not fixable in a workflow** — the definitive control is a **GitHub tag-protection ruleset** (require `v*` tags to be created only via the trusted path / block direct pushes), which is an org-admin setting. The gate + `Cut Signed Release` workflow fully cover the realistic case (a release cut from current `main`, which is what actually slipped on v0.28.0). Recommending the ruleset as the server-side complement. WONTFIX in code. — Co-coded with Jale 🤖

abacigil (approved):

Issue #346, opened 2026-08-28
Discovery run errors: collapse to a small warning icon by default (expand on click) — project home + run detail


jale-dbox:
## Current behavior

When a discovery run errors (e.g. an analysis area fails), the **full error text is shown by default** — a red block:
- **Project home:** `ui/dashboard/src/app/projects/[id]/page.tsx:696` → `{run.error && <div style={{color: var(--db-red-text)}}>{run.error}</div>}`
- **Run detail:** `ui/dashboard/src/app/projects/[id]/discoveries/[runId]/page.tsx:152` → the red *"N area(s) failed during analysis"* banner + the full raw error (see screenshot: a wall of `bedrock/openai-compat … ValidationException …` text).

The error **persists across refresh even after Dismiss** (via `dismissedRunId`, `page.tsx:55/83/375`). **We want to keep that persistence** — the point below is only about the *default visibility*, not removing the error.

## Wanted behavior

Do **not** render the full error by default. Instead:
- Show a **small warning/alert icon** (e.g. ⚠️) next to the run status/title when the run has an error or is `Partial`.
- **Clicking the icon expands** the full error (popover or inline collapsible). Collapsing hides the detail again; the **icon indicator stays**.
- Same treatment on **both** the **project home** and the **discovery run detail** page.
- Preserve current persistence: the indicator survives refresh; **Dismiss collapses the detail, it does not remove the indicator.**

## Files

- `decisionbox-platform/ui/dashboard/src/app/projects/[id]/page.tsx` (`:696` error div, `:654` Dismiss, `:55/83/375` dismiss ref)
- `decisionbox-platform/ui/dashboard/src/app/projects/[id]/discoveries/[runId]/page.tsx` (`:152` "areas failed during analysis" banner)
- **Enterprise mirror** (same change): `decisionbox-enterprise/ui/src/app/projects/[id]/page.tsx` (`:733`)

## Acceptance criteria

- No full error block by default; a compact warning icon appears whenever `run.error` / `summary.errors` is present.
- Clicking the icon reveals the full error; collapsing hides it; the icon **persists across refresh**.
- Applied consistently on project home + run detail, and mirrored in the enterprise dashboard.
- Long raw errors (like the GLM-5 context-length dump) are shown in the expanded view in a scrollable / wrapped container, not the page body.

jale-dbox:
Jale opened a pull request: https://github.com/decisionbox-io/decisionbox-platform/pull/349

[...]

PR #421, opened 2026-09-17
feat(agent): let exploration ask what has been decided about correlating two datasources

jale-dbox:
Correlating across datasources means hopping — read a bounded set of values out of one, then filter the other by them — and which field binds the two is chosen by the model from **names alone**. Two systems can agree on a name and disagree on everything else, and nothing in the query text says which: an order id that is an order id on both sides looks exactly like an order id on one side and a session id on the other.

Somebody may already have looked at that question and written the answer down. Ask had a seam for asking (`ValidateJoinKey`); discovery had none, so it explored with no access to any of it.

Addresses #346.

## What changed

**A read-only seam.** `agentplugin.RegisterCorrelationProvider` — single registration, panic on duplicate, `recover` around the call, and **nothing wired answers nothing**. States that aren't one of the three recognised ones are dropped rather than passed on: everything downstream renders a state as an instruction, and one nobody recognises is neither "use this" nor "do not".

**A `get_correlations` action.** `{"get_correlations": {"a": "...", "b": "..."}}`, plus the tool-use envelope the Claude/OpenAI-shaped models emit whatever the prompt asks for. Ids are validated against the run's datasources and an unknown one comes back as an observation **naming the valid ones** — a model cannot retry with an id it was never told about. Per-run budget with the remaining count in every answer, the way the schema actions already work.

**The answer is written to be acted on.** A rejection leads, under its own heading, carrying the provider's reason — a model follows "these two do not hold the same values" where it talks itself out of "do not use this". It says it is not a preference, holds in either direction, and closes the escape: the decision is about the field, not about how it is spelled. And the substitution section is never silent, because the failure mode of a bare prohibition is a model reaching for the neighbouring key — a pair with no reviewed alternative is told so outright, and told to report the gap rather than invent one.

**Rejections are named in the routing contract, not only in the answer.** A tool is only consulted if the model calls it, so a prohibition that lived only inside the answer is invisible to a run that never asks. Rejections are few — recording one takes a person deciding to — so they are named inline, bounded by `DISCOVERY_MAX_REJECTED_PAIRINGS`, with an overflow line pointing at the tool for the rest. Dropping a prohibition silently would be the worst thing this block could do. Confirmed and declared keys stay in the tool's answer: those are preferences, and there can be many.

**The action is taught in the routing contract, not in the domain packs.** The pack prompts are customisable per project, so an action taught there would change the prompt **every** project receives and still silently miss every project that has customised one — which is the population most likely to have several datasources. Taught in the multi-warehouse section it reaches exactly the runs that can use it. The worked call names two of *this run's* datasources rather than invented ids.

**Telemetry.** A run-level `correlation_lookup_calls` counter, and the action gets its own dashboard step type — the classifier's default arm renders anything it doesn't recognise as a *query* step, so a step that ran no SQL would have been shown as one that did.

## Nothing moves when nothing has been decided

No decisions, nothing wired, or a single-datasource run → no block, no tool mention, and the multi-warehouse prompt is **byte-identical**. The existing golden (`datasources_prompt_all_sql.golden`) passes untouched, which is the proof rather than a chore.

Silence stays neutral in the answer too. Nobody having reviewed a pair is not evidence that no key exists, so an empty answer says exactly that — otherwise every project that has never reviewed a pairing quietly gets more timid than it is today. A read that **failed** is kept distinct from an empty one for the same reason pointed the other way: a read that did not happen has not established that a pairing is unreviewed.

## Advisory, and the wording is not

Nothing here refuses a query. A run that ignores the contract, never calls the action and hops on a rejected pairing produces the wrong correlation and is not stopped. The deterministic refusal is separate work; what the instructions can do, they do. Saying so here so the block above isn't read as a guarantee it isn't.

## Verification

- `go test ./...` green across `libs/go-common`, `services/agent`, `services/api`. One failure in the full agent sweep is environmental and pre-existing — `TestLoad_MissingMongoDBURI` passes with `env -u MONGODB_URI` (the dev stack exports it).
- `golangci-lint run ./...` clean on all three modules; `gofmt` clean on every file this diff touches.
- `scripts/lint-docs.sh` clean on `docs/` and on the CHANGELOG entry.
- `npx tsc --noEmit` clean on the dashboard (one added optional field).
- New tests: seam registry (unwired, passthrough, panic→error, duplicate panics, unrecognised state dropped, reason fallback); parser in both shapes, through a `<think>`-preamble, and the probe itself; the answer's wording asserted as behaviour — prohibition ordering, the reason, the "not a preference" line, the spelling escape, the explicit "nothing" alternative, the neutral empty case, and a failed read not borrowing the empty wording; unknown id, same id twice, missing pair, unwired lookup, budget exhaustion; the contract block's presence, absence, placement, cap and overflow; the stopping rule's subject test and the step classifier.

**One thing I did not test.** The counter's Mongo field name (`correlation_lookup_calls`) is a two-line mirror of the two beside it, checked by inspection — the repository test that would exercise it is behind the `integration` tag, and that lane is not part of this repo's CI. The routing that decides *which* actions reach the incrementer sits on a concrete `*RunRepository` with no seam to fake, and adding one for this would be a wider change than the feature.

Part of #252

— Co-coded with Jale 🤖


jale-dbox:
## Final report

**What.** The exploration agent can now ask what has been decided about correlating two of a run's datasources, and is told outright about the pairings somebody rejected. A generic seam (`agentplugin.RegisterCorrelationProvider`), a `get_correlations` action, and a block in the multi-warehouse routing contract that teaches the action and names the prohibitions.

**Why.** Which field binds two datasources was chosen by the model from names alone, with nothing to consult. Two systems can agree on a name and disagree on everything else, and that is how a confidently wrong correlation reaches an insight.

**How.** Detail is in the PR body; the shape of it is: the seam answers nothing when nothing is wired, the action is offered only on runs whose contract taught it, and a rejection is stated as a prohibition with its reason rather than as a row in a list — because a tool is only consulted if the model calls it.

## Verification

- **9 Codex rounds, 7 findings, all valid, all fixed.** Round 8 was clean; round 9 re-ran after a post-clean test commit, because a clean verdict only covers the diff it saw. What they caught, in order:
  1. The record-grain restriction was stated unconditionally, so on a 3-datasource run one rejection on A↔B told the model not to correlate B↔C record-by-record either — a prohibition nobody recorded, and the opposite of what the action answers for an unreviewed pair.
  2. The lookup was wired for every multi-datasource run, so deployments with **no provider at all** had their opening message changed to advertise an action that could only answer "nothing".
  3. A startup read that **failed** returned the same empty slice as a project with nothing reviewed — a decision store blip would have silently dropped every rejection for a whole run.
  4. The parse-retry nudge listed four actions, steering a run off the fifth at exactly the moment it had just tried it. (This overturned a call I made deliberately in the plan; its argument was better than mine.)
  5. A negative `MaxCorrelationLookupsPerRun` is documented as *off* and clamped to zero — and the guard was `max > 0 && …`, so the one value that disables the action was the one value that never stopped it.
  6. Then the same predicate in two places disagreed again (initial message fixed, repair nudge not). Now one `correlationsOffered()`.
  7. Unbounded provider calls: the error path degrades a failed read, but a provider that **hangs** never fails — it would have held the run for the whole run context, which defaults to 24h.
  8. A run that degraded to one routable datasource was still offered the action, with a worked call naming that datasource twice; and then still paid the startup read for a result nobody would use.
- **Mutation-tested, 14/14 caught.** Each new assertion was proven to fail against a deliberately broken implementation: the "not a preference" line removed, the empty answer reworded to claim no key exists, a failed read made to borrow the empty wording, the no-alternative branch silenced, the budget guard reverted, the inline rejections suppressed, the overflow line dropped, the restriction un-scoped, `correlatable` weakened, `unread` collapsed to empty, the timeout removed, the extractor's action key deleted, and the contract made to render on an empty set. One survivor was real and is now fixed: the repair-nudge test covered the function but not its **call site**, which is exactly where that predicate had already drifted once — it is now driven through `runStepWithRetry` with an unparseable response.
- `go test ./...` green across `libs/go-common`, `services/agent`, `services/api`; `golangci-lint` clean on all three; `gofmt` clean on every file touched; `scripts/lint-docs.sh` clean; `npx tsc --noEmit` clean on the dashboard. CI green on `d5a5be4`.
- One pre-existing environmental failure in the full agent sweep: `TestLoad_MissingMongoDBURI` passes under `env -u MONGODB_URI` (the dev stack exports it).
- No Copilot review appeared — auto-review is not active on this repository (consistent with #409, #410, #411, #415, none of which carry a bot review). Stating it rather than waiting on it.

## Known limitations

- **This is steering, not enforcement.** A run that ignores the contract and never calls the action still produces the wrong correlation. Nothing here refuses a query; the deterministic gate is separate work.
- **`lookup_schema` and `search_tables` carry the same budget defect** finding 5 fixed here — a negative budget is documented as off and behaves as unlimited. Deliberately left alone: changing what those two do with a budget is a behaviour change to existing actions rather than part of this, and nothing currently sets a negative budget for either. Worth its own issue.
- The run counter's Mongo field name is a two-line mirror of the two beside it, checked by inspection — the repository test that would exercise it is behind the `integration` tag, and that lane is not part of this repo's CI.

— Co-coded with Jale 🤖

